PR #9, opened 2017-09-04
BCN -Byron & Jordi

ByronGBP:
@jviles

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,45 @@
-var SortedList = function() {};
+var SortedList = function() {
+  this.length = 0;
+  this.list = [];
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+  this.list.sort(function(a,b){
+    if (a > b){return 1;}
+    else if (a < b){return -1;}
+    else{return 0;}
+  });
+  this.length++;
+};
+
+SortedList.prototype.get     = function(pos) {
+  return this.list[--pos];
```
tawebbcn: Be careful, you are mutating pos!

starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+SortedList.prototype.max = function() {
+    return this.list[this.length - 1];
+};
+
+SortedList.prototype.min = function() {
+  return this.list[0];
+};
+SortedList.prototype.average = function() {
+
+  var total = this.list.reduce(function(sum, value){
```
tawebbcn: Can you reuse another function to do the sum?

tawebbcn:
Good job! :+1:

PR #11, opened 2017-09-04
BCN - Mark and Tania

markvertra:
@taniabatista

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,37 @@
-var SortedList = function() {};
-
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,37 @@
...
+    return this.list[0];
+  };
+  SortedList.prototype.average = function() {
+    return this.sum() / this.length;
+  };
+  SortedList.prototype.sum     = function() {
+    function reduceFn(total, curr) {
+      return total + curr;
+    }
+      return this.list.reduce(reduceFn, 0);
```
tawebbcn: You don't need to declare another function (reduceFn), but it's ok.

tawebbcn:
Good job! :+1:

PR #12, opened 2017-09-04
BCN Will + Andrea

willjthorp:


tawebbcn:
Hi Andrea and Will,
There are a couple of things you could improve in the add and the max prototype functions:

- The add could be returned without and if/else with this line: 
   **return a - b;**

- Using the pop in the max  could be tricky with its results you could try and return: 
  **return this.list[this.length-1];**

Keep pushing!

PR #13, opened 2017-09-04
BCN - Gabriele & Markus

MHFP:
@digitalele

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,110 @@
-var SortedList = function() {};
+class SortedList {
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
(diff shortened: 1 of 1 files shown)


tawebbcn:
Hi Markus and Gabriele! 

There are a couple of things you could improve in the add and the max prototypes functions:

- The add could be returned without an if/else with this line: 
   **return a - b;**

- The max could be returned with this : 
  **return this.list[this.length-1];**

- The min with this : 
    **return this.list[0];**

- The average with this :
  **return this.sum()/this.length;**

- The sum with this one:
  **var sum = 0;
  for (var i=0; i<this.length; i++){
   sum += this.list[i];
  }
  return sum;**

All the tests are working and you made a great effort to solve this, **GOOD JOB!** 

Keep pushing!

P.s. Remove the uncommented code so you’ll have a cleaner exercise in GitHub :)

PR #32, opened 2017-11-06
BCN - Juan Cuadra & Felice Pozzi 

jcuadrad:


starter-code/index.js
```diff
@@ -1,10 +1,47 @@
...
+SortedList.prototype.add = function(item) {
+  this.list.push(item);
+  this.length++;
+  if (this.length > 1) {
+    this.list.sort(function(a, b) {
+      return a - b;
+    });
+  }
+};
+SortedList.prototype.get = function(pos) {
```
tawebbcn: How about checking if pos is in the boundaries of the array?

starter-code/index.js
```diff
@@ -1,10 +1,47 @@
...
+};
+SortedList.prototype.max = function() {
+  return Math.max.apply(null, this.list);
+};
+SortedList.prototype.min = function() {
+  return Math.min.apply(null, this.list);
+};
+SortedList.prototype.average = function() {
+  if (this.length >= 1) {
+    var result = this.list.reduce(function(accumulator, currentValue) {
```

PR #33, opened 2017-11-06
BCN - Desiree and Dominik

dlasch089:
@dvillool

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,49 @@
-var SortedList = function() {};
+var SortedList = function() {
+  this.list = [];
+  this.length = 0;
 
-SortedList.prototype.add     = function(item) {}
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,49 @@
...
+  this.length++;
+  if (this.length > 1) {
+    this.list.sort(function(a, b) {
+      return a - b;
+    });
+  }
+};
+
+
+SortedList.prototype.get = function(pos) {
```

starter-code/index.js
```diff
@@ -1,10 +1,49 @@
...
+  return this.list[this.length - 1];
+};
+
+
+SortedList.prototype.min = function() {
+  return this.list[0];
+};
+SortedList.prototype.average = function() {
+  var sum = 0;
+  for (var ix = 0; ix < this.length; ix++) {
```

PR #34, opened 2017-11-06
BCN - Jean & Cristian

jnothomb:
@cristiandiazz

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,50 @@
-var SortedList = function() {};
+var SortedList = function() {
+  this.list = [];
+  this.length = 0;
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+
+SortedList.prototype.add = function(item) {
+  this.list.push(item);
+  this.list.sort(function(a, b) {
+    return a - b;
+  });
+  this.length++;
+};
+
+SortedList.prototype.get = function(pos) {
```

starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+  return this.list[pos - 1];
+};
+
+
+SortedList.prototype.max = function() {
+  return this.list[this.list.length - 1];
```
tawebbcn: You can acces the length with this.length (it is a property).

starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+};
+
+SortedList.prototype.min = function() {
+  return this.list[0];
+};
+
+SortedList.prototype.average = function() {
+  var total = 0;
+  for (var ix = 0; ix < this.list.length; ix++) {
```

PR #35, opened 2017-11-06
BCN - Clara Ameller - Sarah Fontes

ClaraAmeller:
@sarahfontes


tawebbcn:
Are you sure you have delivered the proper version? There is nothing done. :confused:

PR #36, opened 2017-11-06
@BCN Lucas + Cristina + Constantinos

lacazeto:
@crismoreno
@cyenis

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,7 +1,26 @@
-var SortedList = function() {};
+var SortedList = function() {
+  this.length = 0;
+  this.list = [];
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,29 @@
...
+SortedList.prototype.add     = function(item) {
+    this.list.push(item);
+    this.length++;
+    this.list.sort(function(a,b){
+        return a -b;
+    });
+    }
+}
+SortedList.prototype.get     = function(pos) {
+    if(pos < this.list.length){
```
tawebbcn: Be careful, it would be out of bounds if pos is 0 or a negative number.

starter-code/index.js
```diff
@@ -1,10 +1,29 @@
...
+        return this.list[pos];
```
tawebbcn: You want to return the pos - 1 position. E.g.: first element (index 0) if the user input position 1.

PR #37, opened 2017-11-06
BCN - Torgeir & Eloi

torgeirbeyer:
@eloigil


starter-code/index.js
```diff
@@ -1,10 +1,57 @@
...
+
+
+SortedList.prototype.add     = function(item) {
+    this.list.push(item);
+    this.list = this.list.sort(function(a,b) {
+        return a - b;
+    });
+    this.length++; 
+}
+SortedList.prototype.get     = function(pos) {
```
tawebbcn: What about checking if pos is into the boundaries of the array?

starter-code/index.js
```diff
@@ -1,10 +1,57 @@
...
+    return this.list[pos - 1];
+}
+SortedList.prototype.max     = function() {
```
tawebbcn: Good job, but there are Math methods and, even more, the array is sorted.

starter-code/index.js
```diff
@@ -1,10 +1,57 @@
...
+        if(minNum === null){
+            minNum = this.list[i];
+        }
+        if(this.list[i] < minNum){
+            minNum = this.list[i];
+        }
+    } return minNum;
+}
+SortedList.prototype.average = function() {
+    var numAverage = null;
```
tawebbcn: Better start it to 0.

starter-code/index.js
```diff
@@ -1,10 +1,57 @@
...
+    for(var i = 0; i < this.list.length; i++){
```

PR #38, opened 2017-11-06
BCN PP TIAGO ELVIN

3lv27:
@tiago1999

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,46 @@
-var SortedList = function() {};
+var SortedList = function() {
+  this.list = [];
+  this.length = 0;
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+  if (this.length > 1) {
+    this.list.sort(function(a, b) {
+      return a - b;
+    });
+  }
+
+
+};
+
+SortedList.prototype.get = function(pos) {
```
tawebbcn: What about checking if pos is into the boundaries of the array?

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+
+SortedList.prototype.max = function() {
+  return this.list[this.list.length - 1];
+};
+SortedList.prototype.min = function() {
+  return this.list[0];
+};
+SortedList.prototype.average = function() {
+  var acum = 0;
+  for (var i = 0; i < this.list.length; i++) {
```

PR #39, opened 2017-11-06
BCN Jordi y David

joordiii:
@otidlam

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,54 @@
-var SortedList = function() {};
+var SortedList = function() {
+  this.list = [];
+  this.length = 0;
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,54 @@
...
+};
+
+SortedList.prototype.add = function(item) {
+  this.list.push(item);
+  this.length++;
+  this.list = this.list.sort(function(a, b) {
+    return a - b;
+  });
+};
+SortedList.prototype.get = function(pos) {
```

starter-code/index.js
```diff
@@ -1,10 +1,54 @@
...
+  return this.list[pos - 1];
+};
+SortedList.prototype.max = function() {
```
tawebbcn: Good job, but there are Math methods and, even more, the array is sorted.

starter-code/index.js
```diff
@@ -1,10 +1,54 @@
...
+  for (var ix = 0; ix < this.list.length; ix++) {
+    if (minValue > this.list[ix]) {
+      minValue = this.list[ix];
+    }
+  }
+  return minValue;
+};
+SortedList.prototype.average = function() {
+  var average = 0;
+  for (var ix = 0; ix < this.list.length; ix++) {
```

PR #42, opened 2017-11-06
BCN - Clara Ameller - Sarah Fontes

ClaraAmeller:
@sarahfontes

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,50 @@
-var SortedList = function() {};
+var SortedList = function() {
+    this.length = [];
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add     = function(item) {
+    this.length.push(item);
+    this.length.sort();
+};
+
+SortedList.prototype.size     = function() {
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,50 @@
-var SortedList = function() {};
-
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+var SortedList = function() {
+    this.length = [];
```
tawebbcn: Maybe is better to have two properties:
```
this.list = [];
this.length = 0;
```

starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+    this.length.push(item);
+    this.length.sort();
+};
+
+SortedList.prototype.size     = function() {
+    return this.length.length;
+};
+
+SortedList.prototype.get     = function(pos) {
+    this.length.sort();
```
tawebbcn: The list is already sorted!

starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+
+SortedList.prototype.min     = function() {
+    return Math.min.apply(Math, this.length);
+};
+
+SortedList.prototype.average = function() {
+    if (this.length.length === 0) {
+        return 'EmptySortedList';
+    }
+    return (this.length.reduce((prev, next) => prev + next) / this.length.length);
```

PR #51, opened 2017-12-09
PAR Charlotte node.js intro DONE

Charlottebtmn:


starter-code/index.js
```diff
@@ -1,10 +1,47 @@
...
+  if (this.items.length===0) {
+    this.items.push(item);
+  }
+  else {
+    var i=0;
+    while (item>this.items[i]) {
+      i++;
+    }
+    // console.log(this.items, i);
+    this.items.splice(i, 0, item);
```
ta-web-paris: well done ! I think it's might be even better (faster) than the solution we thought of ;)

PR #81, opened 2018-01-29
BCN - Rodrigo and Pedro

rodrigodaoud:
@pedrodiogoneto

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,38 @@
-var SortedList = function() {};
+class SortedList {
+  constructor () {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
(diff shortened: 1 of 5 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+class SortedList {
+  constructor () {
+    this.items = [];
+    this.length = this.items.length;
+  }
+
+  add (item) {
+    this.items.push(item);
+    this.items.sort((a, b) => (a - b));
+    this.length = this.items.length;
```
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+  }
+
+  get (pos) {
+    return this.items[pos - 1];
```

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+    return Math.min(...this.items);
+  }
+
+  average () {
+    return this.sum() / this.items.length;
+  }
+
+  sum () {
+    let sum = 0;
+    for (let i = 0; i < this.items.length; i++) {
```
tawebbcn: Good job, you can also use a reduce.

PR #82, opened 2018-01-29
BCN / Oscar y David

oscarjorgejorge:
@davidalexteixeira


starter-code/index.js
```diff
@@ -1,10 +1,54 @@
-var SortedList = function() {};
+var SortedList = function() {
+    this.items = [];
+    this.length = this.items.length;
```
```
this.length = 0;
```

starter-code/index.js
```diff
@@ -1,10 +1,54 @@
...
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
+
+SortedList.prototype.add     = function(item) {
+    this.items.push(item);
+    this.length = this.items.length;
```
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,54 @@
...
+    this.items.sort((a,b) => {
+        return a-b;
+    }) 
+}
+
+SortedList.prototype.get     = function(pos) {
+    return this.items[pos - 1];
```

PR #83, opened 2018-01-29
BCN - James and Sheng

jsavory-94:
@theRealSheng

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,42 @@
-var SortedList = function() {};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+// SortedList.prototype.add = function (item) {};
+// SortedList.prototype.get = function (pos) {};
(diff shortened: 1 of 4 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,42 @@
...
+// SortedList.prototype.get = function (pos) {};
+// SortedList.prototype.max = function () {};
+// SortedList.prototype.min = function () {};
+// SortedList.prototype.average = function () {};
+// SortedList.prototype.sum = function () {};
+
+class SortedList {
+  constructor () {
+    this.items = [];
+    this.length = this.items.length;
```
```
this.length = 0;
```

starter-code/index.js
```diff
@@ -1,10 +1,42 @@
...
+  };
+
+  add (x) {
+    this.items.push(x);
+    this.length = this.items.length;
```
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,42 @@
...
+  max () {
+    return Math.max(...this.items);
+  };
+
+  min () {
+    return Math.min(...this.items);
+  };
+
+  average () {
+    return this.items.length === 0 ? [] : this.items.reduce((acc, items) => acc + items) / this.items.length;
```
tawebbcn: Good job with the reduce in sum, but here better reuse your sum function:
```
return this.items.length === 0 ? [] : this.sum() / this.items.length; 
```

PR #84, opened 2018-01-29
BCN - DANIELLE & LUIS

luisacherrera:
@danielleleb

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,34 @@
-var SortedList = function() {};
+var SortedList = function () {
+  var self = this;
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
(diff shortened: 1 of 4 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,34 @@
...
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+var SortedList = function () {
+  var self = this;
+
+  self.items = [];
+  self.length = self.items.length;
```
```
this.length = 0;
```

starter-code/index.js
```diff
@@ -1,10 +1,34 @@
...
+};
+
+SortedList.prototype.add = function (item) {
+  var self = this;
+
+  self.items.push(item);
+  self.length = self.items.length;
```
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,34 @@
...
+
+  self.items.sort(function (a, b) {
+    return a - b;
+  });
+};
+
+SortedList.prototype.get = function (pos) {
+  var self = this;
+  return self.items[pos - 1];
```

PR #85, opened 2018-01-29
tBCN Alla and Vlad - to easy 4 us:)

vladcio:
@allarom

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,18 @@
-var SortedList = function() {};
+// / SortedList.prototype.get = function (pos) {};
+// SortedList.prototype.max = function () {};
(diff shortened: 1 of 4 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,18 @@
...
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+class SortedList {
+  constructor () {
+    this.items = [];
+    this.length = this.items.length;
```
```
this.length = 0;
```

PR #86, opened 2018-01-29
BCN - Jana & Josep

Jsalvat:
@jjanastinn

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,32 @@
-var SortedList = function() {};
+var SortedList = function () {
+  this.items = [];
+  this.length = this.items.length;
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add = function (item) {
+  this.items.push(item);
+  this.length = this.items.length;
+
+  this.items.sort(function (a, b) {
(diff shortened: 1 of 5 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,32 @@
-var SortedList = function() {};
+var SortedList = function () {
+  this.items = [];
+  this.length = this.items.length;
```
```
this.length = 0;
```

starter-code/index.js
```diff
@@ -1,10 +1,32 @@
...
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add = function (item) {
+  this.items.push(item);
+  this.length = this.items.length;
```
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,32 @@
...
+
+  this.items.sort(function (a, b) {
+    return a - b;
+  });
+};
+SortedList.prototype.get = function (pos) {
+  return this.items[pos - 1];
```

starter-code/index.js
```diff
@@ -1,10 +1,32 @@
...
+};
+SortedList.prototype.max = function () {
+  return Math.max(...this.items);
+};
+SortedList.prototype.min = function () {
+  return Math.min(...this.items);
+};
+SortedList.prototype.average = function () {
+  var total = 0;
+  for (var i = 0; i < this.items.length; i++) {
```
tawebbcn: This code can do inside the method sum and then you can reuse it.

PR #87, opened 2018-01-29
BCN - Santi & Stephie & Pere

pereamoros:
@santironhacker @kissy89

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,53 @@
-var SortedList = function() {};
+// var SortedList = function () {};
+// SortedList.prototype.add = function (item) {};
+// SortedList.prototype.get = function (pos) {};
+// SortedList.prototype.max = function () {};
+// SortedList.prototype.min = function () {};
+// SortedList.prototype.average = function () {};
+// SortedList.prototype.sum = function () {};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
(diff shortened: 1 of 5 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+class SortedList {
+  constructor () {
+    this.items = [];
+    this.length = this.items.length;
```
```
this.length = 0;
```

starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+  }
+
+  add (item) {
+    this.items.push(item);
+    this.items.sort((a, b) => a - b);
+    this.length = this.items.length;
```
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+  }
+
+  get (i) {
+    if (i <= this.length && i >= 0) {
```
tawebbcn: Be careful! i can't be equal to 0 (then it would try to return the index -1).

starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+
+  min () {
+    return Math.min(...this.items);
+  }
+
+  average () {
+    if (this.length === 0) {
+      return this.items;
+    } else {
+      const sum = this.items.reduce((acc, currentValue) => acc + currentValue);
```
tawebbcn: Good job with the reduce in sum, but here better reuse your sum function:
```
const sum = this.sum(); 
```

PR #88, opened 2018-01-29
BCN FILIPE - DANI

DTDK:
@NCODER92

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,3 +1,5 @@
+/*
+
 var SortedList = function() {};
 
 SortedList.prototype.add     = function(item) {}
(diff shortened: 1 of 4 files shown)


starter-code/index.js
```diff
@@ -6,5 +8,52 @@ SortedList.prototype.max     = function() {}
...
+class SortedList{
+  constructor (){
+    this.item = [];
+    this.length = 0;
+
+  }
+
+  add(item){
+    this.item.push(item);
+    this.length = this.item.length;
```
tawebbcn: This will work, but better:
```
this.length++;
```

starter-code/index.js
```diff
@@ -6,5 +8,52 @@ SortedList.prototype.max     = function() {}
...
+    this.item.sort((a, b) => a - b);
+  }
+
+  get(pos) {
+    
+     return this.item[pos-1];
```

starter-code/index.js
```diff
@@ -6,5 +8,52 @@ SortedList.prototype.max     = function() {}
...
+  max(){
+    return Math.max(...this.item);
+  }
+
+  min(){
+    return Math.min(...this.item);
+  }
+
+  average(){
+    return this.sum()/this.item.length;
```
tawebbcn: Good job reusing your function.

starter-code/index.js
```diff
@@ -6,5 +8,52 @@ SortedList.prototype.max     = function() {}
...
+  }
+  
+  sum(){
+    if(this.length === 0){
+      return "EmptySortedList";
+    }
+    return this.item.reduce((accumulator, currentValue) => accumulator + currentValue);
```
tawebbcn: Good job using a reduce.

PR #89, opened 2018-01-29
BCN Linda & Martin

martinX308:
@lindaluisa

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,49 @@
-var SortedList = function() {};
+// var SortedList = function() {};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+// SortedList.prototype.add     = function(item) {}
+// SortedList.prototype.get     = function(pos) {}
+// SortedList.prototype.max     = function() {}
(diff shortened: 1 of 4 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,49 @@
...
+class SortedList {
+  constructor () {
+    this.items = [];
+    this.length = 0;
+  }
+
+  add (item) {
+    this.items.push(item);
+    this.items.sort((a, b) => a - b);
+    this.length = this.items.length;
```
tawebbcn: This will work, but better:
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,49 @@
...
+  }
+
+  get (position) {
+    return this.items[position - 1];
```

starter-code/index.js
```diff
@@ -1,10 +1,49 @@
...
+  max () {
+    return Math.max(...this.items);
+  }
+
+  min () {
+    return Math.min(...this.items);
+  }
+
+  sum () {
+    return this.items.reduce((acc, item) => acc + item, 0);
```
tawebbcn: Good job using a reduce.

starter-code/index.js
```diff
@@ -1,10 +1,49 @@
...
+  }
+
+  //   sum () {
+  //     return this.items.reduce(function (acc, item) {
+  //       return acc + item;
+  //     }, 0);
+  //   }
+
+  average () {
+    return this.sum() / this.items.length;
```
tawebbcn: Good job reusing your function.

PR #90, opened 2018-01-29
BCN - Matt & Victor - Node JS PP

victorpellerin:
@mattmegarry

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,27 @@
-var SortedList = function() {};
+// var SortedList = function() {};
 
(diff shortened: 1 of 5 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,27 @@
...
+// SortedList.prototype.min     = function() {}
+// SortedList.prototype.average = function() {}
+// SortedList.prototype.sum     = function() {}
+
+// const array = [1, 7, 10, 37, 98];
+
+class SortedList {
+  constructor () {
+    this.items = [];
+    this.length = this.items.length;
```
```
this.length = 0;
```

PR #95, opened 2018-02-12
[MIA][Scott-Rowell]

rowellscott:


starter-code/index.js
```diff
@@ -1,10 +1,29 @@
...
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add     = function(item) {
+    this.items.push(item);
+    this.items.sort(function(a, b){return a-b})
+    this.length = this.items.length
+}
+SortedList.prototype.get     = function(pos) {
+    return this.items[pos-1];
```
sandrabosk: What if you ask for position of an element on position, lets say, 6th and you have less elements? You have let your user know that the element on that position doesn't exist. So before this return you would put something like this:
```
        throw Error("OutOfBounds");
    }
```

sandrabosk:
Good job Scott. Thank you for committing.

starter-code/index.js
```diff
@@ -1,10 +1,29 @@
...
+}
+SortedList.prototype.max     = function() {
```
```
 if (this.length === 0) {
      throw Error("EmptySortedList");
    }
```

starter-code/index.js
```diff
@@ -1,10 +1,29 @@
...
+  return Math.max(...this.items)
+}
+SortedList.prototype.min     = function() {
```
```
 if (this.length === 0) {
      throw Error("EmptySortedList");
    }
```

PR #96, opened 2018-02-12
[MIA] [Stephen - Jackson]

stephjackson:
Please fix the array position thing.

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,50 @@
-var SortedList = function() {};
+var SortedList = function() {
+  this.items = [];
+  this.length = this.items.length;
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+};
+
+SortedList.prototype.add     = function(item) {
+  this.items.push(item);
+
+  this.items.sort(function(a,b){
+    return (a-b);
+  });
+
+  this.length = this.items.length;
```
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+}
+SortedList.prototype.get     = function(pos) {
+  //WHY IS THIS LIKE THIS WHO DOES THIS WHAT MONSTER DID THIS
```
sandrabosk: We have to check the array has any elements:
```
        throw Error("OutOfBounds");
    }
```

sandrabosk:
I laughed at your comment :) I felt the same when I was failing the test like 10 times.
Good job and thank you for submitting.

PR #97, opened 2018-02-12
[MIA] [Jonathan Sanchez]

jonathanlsanch:


starter-code/index.js
```diff
@@ -1,10 +1,48 @@
...
+};
+
+SortedList.prototype.add     = function(item) {
+    this.items.push(item);
+
+    this.items = this.items.sort(function(a,b) {
+        return (a-b);
+    })
+
+    this.length = this.items.length;
```
sandrabosk: Since we pushed into array, the length of the array is incrementing:
```
    this.length ++;
```

starter-code/index.js
```diff
@@ -1,10 +1,48 @@
...
+}
+SortedList.prototype.get     = function(pos) {
+    return this.items[pos - 1];
```
sandrabosk: What if you ask for position of an element on position, lets say, 6th and you have less elements? You have let your user know that the element on that position doesn't exist. So before this return you would put something like this:
```
        throw Error("OutOfBounds");
    }
```

sandrabosk:
Good job! Thank you for submitting.

PR #98, opened 2018-02-12
[MIA - Pablo Gomez]

ZTylerDurden:
Started off slow and then got help from Torre and his pair group.

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,36 @@
-var SortedList = function() {};
+var SortedList = function() {
+this.items = [];
+this.length = 0;
+};
 
-SortedList.prototype.add     = function(item) {}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,39 @@
...
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add     = function(item) {
+    this.items.push(item)
+    this.items.sort(function(a,b) {
+        return a - b
+    })
+    this.length = this.items.length
```
sandrabosk: Since we are pushing into array, the length of the array is incrementing:
```
this.length++;
```

PR #101, opened 2018-02-12
[MIA] [Jeremy Fox - node intro exercise]

jfoxxxx:


starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+    this.items = [];
+    this.length = 0;
+ };
+    
+ SortedList.prototype.add     = function(item) {
+    this.items.push(item);
+    this.items.sort((a,b)=>{
+        return a - b;
+      });
+    this.length = this.items.length;
```
sandrabosk: Since we are pushing into array, we should add to the length, so the proper code would be:
```
this.length++;
```

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+ };
+ 
+ SortedList.prototype.get     = function(pos) {
+    if(this.length === 0)
```
sandrabosk: What if you ask for position of an element on position, lets say, the 6th and you have less elements? You have let your user know that the element on that position doesn't exist. So before this return you would put something like this:
```
        throw Error("OutOfBounds");
    }
```

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+        return;
+    return this.items[pos - 1];
+ }
+ SortedList.prototype.max     = function() {
+    if(this.length === 0)
```
``` 
if (this.length === 0){
        throw Error("EmptySortedList");
    }
```

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+        return;
+    return this.items[this.length - 1];
+ }
+ SortedList.prototype.min     = function() {
+    if(this.length === 0)
```
sandrabosk: The same approach as the previous one.

sandrabosk:
You showed basic understanding of the Node concepts. Good job so far and keep it that way. Thank you for submitting.

PR #105, opened 2018-02-12
[MIA]Ana Raballo - Inital commit

AnaRaballo:


starter-code/index.js
```diff
@@ -1,10 +1,36 @@
...
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add     = function(item) {
+    this.items.push(item);
+    this.items.sort((a,b) => {
+        return a-b;
+    });
+     this.length++;
+};
+
+SortedList.prototype.get     = function(pos) {
```
sandrabosk: what if you ask for position of an element on position, lets say, 6th and you have less elements? you have let your user know that the element on that position doesn't exist. so before this return you would put something like this:
```
        throw Error("OutOfBounds");
    }
```

sandrabosk:
Good job Ana!

PR #106, opened 2018-02-12
[MIA] Lexy Vanderford

LexyV:


starter-code/index.js
```diff
@@ -1,10 +1,33 @@
...
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add     = function(item) {
+    this.items.push(item);
+    this.items.sort((a, b) => {
+        return a - b;
+    });
+    this.length++;
+};
+SortedList.prototype.get     = function(pos) {
```
sandrabosk: We have to check if we have elements in the array and also how long our array is so we know the "scope" of the array: 
```
        throw Error("OutOfBounds");
    }
```

sandrabosk:
Good job!

PR #131, opened 2018-04-09
PAR Sami & J-B

Sami-Az:


starter-code/index.js
```diff
@@ -1,10 +1,49 @@
...
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add  = function (x) {
+  
+  this.items.push(x);
+  this.items = this.items.sort();
```
ta-web-paris: Unfortunately, this will not sort an array of numbers correctly in all cases, because the default sort is according to unicode and not by size, so 100 comes before 90 for example. You should provide the sort method with your own sort function:
`this.items.sort(function(number1, number2) { return number1 - number 2})`

PR #139, opened 2018-04-09
BCN - Jordi & Rapha

jordicalzada:
@raphamontenegro

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,46 @@
-var SortedList = function() {};
+var SortedList = function() {
+    this.items = [];
+    this.length = this.items.length;
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+SortedList.prototype.add = function(item) {
+    this.items.push(item);
+    this.length += item;
+};
+SortedList.prototype.get = function(pos) {
+    this.items.length
+    return this.items[pos-1];
+
+};
+SortedList.prototype.max = function() {
+    this.items.sort(function (a,b) {
+        return a + b});
+
+    return this.items[0];
+
+
+};
+SortedList.prototype.min = function() {
+    this.items.sort(function(a, b) {
+      return a - b;
+    });
+
+    return this.items[0];
+
+};
+SortedList.prototype.average = function() {
+    var self = this;
+    var sum = 0;
+    self.items.reduce(function(acc, current) {
+      sum = (acc + current);
+    }, 0);
+    return sum / self.items.length;
+};
+SortedList.prototype.sum = function() {
+    var self = this;
+    self.items.reduce(function(acc, current){
+        return acc + current;
+    },0);
+};
 
 module.exports = SortedList;


starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+var SortedList = function() {
+    this.items = [];
+    this.length = this.items.length;
+};
+
+SortedList.prototype.add = function(item) {
+    this.items.push(item);
+    this.length += item;
```
tawebbcn: You have to update the length, not do the sum of the values
```
this.length = this.items.length;
```

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+};
+SortedList.prototype.get = function(pos) {
+    this.items.length
```
tawebbcn: don't need that

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+    return this.items[pos-1];
```
tawebbcn: this is ok! :+1: (I WAS WRONG! when reviewing in class) after checking a bit more the tests, just realised that it compares the values and expects them to be equal.

starter-code/index.js
tawebbcn: You have to sort the array inside this method

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+
+};
+SortedList.prototype.max = function() {
+    this.items.sort(function (a,b) {
```
tawebbcn: (you should sort in add())

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+        return a + b});
+
+    return this.items[0];
```
tawebbcn: :+1:

tawebbcn: you also can use ```Math.max()```

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+
+
+};
+SortedList.prototype.min = function() {
+    this.items.sort(function(a, b) {
+      return a - b;
+    });
+
+    return this.items[0];
+
```
tawebbcn: same as before

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
-var SortedList = function() {};
-
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+var SortedList = function() {
+    this.items = [];
```
tawebbcn: ```var self = this;```

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+    self.items.reduce(function(acc, current) {
+      sum = (acc + current);
+    }, 0);
+    return sum / self.items.length;
+};
+SortedList.prototype.sum = function() {
+    var self = this;
+    self.items.reduce(function(acc, current){
+        return acc + current;
+    },0);
```
tawebbcn:
```
return this.items.reduce(function(sum, ele){
    return sum + ele;
  }, 0);
```

starter-code/index.js
```diff
@@ -1,10 +1,46 @@
...
+
+    return this.items[0];
+
+};
+SortedList.prototype.average = function() {
+    var self = this;
+    var sum = 0;
+    self.items.reduce(function(acc, current) {
+      sum = (acc + current);
+    }, 0);
```
tawebbcn:
```
return (this.sum() / this.length);
```

PR #140, opened 2018-04-09
not done with excersise :(

samanthaferreira:
@fscheugenpflug

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,9 +1,31 @@
-var SortedList = function() {};
+function SortedList(items,length) {
+    this.items = [];
+    this.length = this.items.length;
+};
 
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
+// function get(nth){
+//     var nth;
+//     this.items[nth];
+// }
+
+SortedList.prototype.add = function(item) {
+    this.items.push(item);
+    this.length = this.items.length;
+    this.items.sort(function(a,b){
+        return a > b;
+    });
+}
+SortedList.prototype.get     = function(pos) {
+    this.item[pos];
+    this.length;
+}
+SortedList.prototype.max     = function() {
+    this.items[0];
+}
+SortedList.prototype.min     = function() {
+    var min = this.items.length -1;
+    this.items[min];
+}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+//     var nth;
+//     this.items[nth];
+// }
+
+SortedList.prototype.add = function(item) {
+    this.items.push(item);
+    this.length = this.items.length;
+    this.items.sort(function(a,b){
+        return a > b;
+    });
```
tawebbcn: nice

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+}
+SortedList.prototype.get     = function(pos) {
+    this.item[pos];
```
tawebbcn: ```return this.item[pos - 1];```

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+    this.length;
```
tawebbcn: don't need that

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+}
+SortedList.prototype.max     = function() {
+    this.items[0];
```
tawebbcn: when you sort, the max number becomes the last one``` return this.items[this.items.length - 1];```

tawebbcn: missing a return

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+}
+SortedList.prototype.min     = function() {
+    var min = this.items.length -1;
+    this.items[min];
```
tawebbcn: ```return this.items[0];```

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+}
+SortedList.prototype.average = function() {
+    this.sum = this.items.reduce(add,0);
+    function add(a,b){
```
tawebbcn: you have this method below: ```self.sum()``` it is perfect for doing the sum

starter-code/index.js
```diff
@@ -1,10 +1,38 @@
...
+        return a + b;
+    }
+    this.average = this.sum / this.length;
```
tawebbcn: :+1:

starter-code/index.js
tawebbcn: but reduce is ok

PR #163, opened 2018-04-23
MIA - Neil Rampersad / Jose Arjona - Lab Intro Node

neilxcode:
All test points passed.

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,25 +1,30 @@
+
 var SortedList = function() {
     this.length = 0;
     this.items = [];
 };
+//PUSHES ITEM, SORTS ALL ITEMS IN ARRAY, UPDATES LENGTH PROPERTY//
 SortedList.prototype.add     = function(item) {
     this.items.push(item);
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,10 +1,50 @@
...
+//RETURN Nth ITEM(-1 differenence between NUM & ARRAY)
+SortedList.prototype.get     = function(pos) {
+    return this.items[pos - 1];
+};
+//MAX
+SortedList.prototype.max     = function() {
+    return Math.max.apply(null, this.items);
+}
+//MIN
+SortedList.prototype.min     = function() {
```
ianizaguirre: Very clever using the `.apply` method here. Just so you can compare, here was another way to write this method:
```js
  var minNumber =
    this.list.reduce((minSoFar, number) => {
      return Math.min(minSoFar, number);
    });
  return minNumber;
};
```

ianizaguirre:
Good job, you guys passed all tests.
Keep up the good work! 😁

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #167, opened 2018-04-23
MIA - FTWD - jose arjona, neil rampersad - mocha testing js node lab

josedarjona:


starter-code/index.js
```diff
@@ -1,10 +1,57 @@
...
+return this.items[pos - 1];
+}
+
+SortedList.prototype.max = function() {
+  return Math.max.apply(null, this.items);
+
+}
+
+SortedList.prototype.min = function() {
+  return Math.min.apply(null, this.items);
```
ianizaguirre: Very clever using the `.apply` method here. Just so you can compare, here was another way to write this method:
```js
  var minNumber =
    this.list.reduce((minSoFar, number) => {
      return Math.min(minSoFar, number);
    });
  return minNumber;
};
```

ianizaguirre:
Good job, you guys passed all tests.
Keep up the good work! 😁

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #170, opened 2018-04-23
MIA-JEYSON/ELEYNI-LAB-INTRO-NODE

jmxxii:


starter-code/index.js
```diff
@@ -1,10 +1,45 @@
-var SortedList = function() {};
-
-SortedList.prototype.add     = function(item) {}
-SortedList.prototype.get     = function(pos) {}
-SortedList.prototype.max     = function() {}
-SortedList.prototype.min     = function() {}
-SortedList.prototype.average = function() {}
-SortedList.prototype.sum     = function() {}
+var SortedList = function() {
+  this.items = [],
```
ianizaguirre: Mistake was made here... Object LITERALS end with commas on each line BUT Object Constructors end with Semicolons. You added commas 🧐

starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+SortedList.prototype.add     = function(item) {
+this.items.push(item)
+this.items.sort(function(a,b){
+  return a - b
+})
+this.length ++;
+}
+SortedList.prototype.get     = function(pos) {
+  if(pos < 1 || pos > this.length){
+    throw Error ("OutOfBounds");
```
ianizaguirre: You are not supposed to write any "if" or "throw Error" .... when we look at tests, if you write exactly what the test is "testing" then your not really testing your code but rather just fooling the test into thinking you passed it. 

This was the correct way for this one:
```js
SortedList.prototype.get = function(pos) {
  return this.list[pos - 1];
};
```

starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+  }
+ return this.items[pos-1]
+}
+SortedList.prototype.max     = function() {
+  return Math.max.apply(Math, this.items);
```
ianizaguirre: Clever using the apply method. Another way you could have written it, is using  "reduce":
```js
  var maxNumber =
    this.list.reduce(
      (maxSoFar, number) => {
        return Math.max(maxSoFar, number);
      });
  return maxNumber;
```

starter-code/index.js
```diff
@@ -1,10 +1,45 @@
...
+}
+SortedList.prototype.sum     = function() {
+  var total = 0;
+  for (var i = 0; this.items.length; i++){
+    total += this.items[i];
+  }
+  return total;
+}
+
+var nsl = new SortedList();
```
ianizaguirre: hmm why did you need to write:
```js
var nsl = new SortedList();
nsl.add(1)
nsl.ad
```
???

PR #179, opened 2018-05-28
Ironhack MIA - PTWD - May 2018 - Maximilianus Paredes

XamSP:
Did all of them except the second test, it worked normal on the console.
Seriously wouldn't mind an answer for it.


ta-web-mia:
So, the test is vague lol Its not reading that the length is being added. I added length++ and it worked

PR #189, opened 2018-06-03
 MIA-PTWD-MARIAH-DUNN

teepott99:


ta-web-mia:
Looks good, now the code that really fixes it all belongs to the get. Now the way that the test explains it is "should return element in that position", which is super vague. So heres my hint lol If I put 1 as the parameter it should show me the first element in that array, we know that arrays begin at 0 so how do say that in code?

-Adrian

PR #209, opened 2018-06-18
PAR Nadia-Jihad

eloji9:
@nadiahachemi
We got stuck at the beggining but managed to pass end since we should have made the second part before verifying the code. 
We had a few bumps with "reduce" and forgetting some "{".
It works !

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,10 +1,47 @@
-var SortedList = function() {};
+var SortedList = function() {
+    this.items = [];
+    this.length = this.items.length;
+};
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Good job, you manage to resolve all the errors 💯
I like your code, it's clean and efficient, only one thing is that you should remove the commented code when pushing, so it will be even cleanier 👍 

Hope you like this first exercise with the terminal.. you will have to use it a lot later ;)

If you have any question let us know!

/Arthur

PR #262, opened 2018-07-02
MIA Hannah Loop

hloop:
Almost made it to the finish line (but not quite)


MarcosPalaciosIII:
great work on the node intro lab.  Just remember that you don't have to create another `index.js`  to complete but if you do create another one, make sure that you copy the content from index 2 into index 1

PR #265, opened 2018-07-03
Mia 070218 Yaryna Ravlyk

YokoYolo:


starter-code/index.js
```diff
@@ -1,10 +1,55 @@
...
+class SortedList {
+    constructor (items = []) {
+        this.items = items;
+        this.length = this.items.length;
+    }
+  
+  add(x) {
+    let i = 0;
+    
+    while (x > this.items[i] && i < this.length) {
```
sandrabosk: Loop can be easily replaced with .push()

starter-code/index.js
```diff
@@ -1,10 +1,55 @@
...
+        }
+    this.length++;
+    this.items.splice(i, 0, x);
+  }
+        
+  get(x) {
+    return this.items[x - 1];
+  }
+            
+  max() {
```
sandrabosk: We can have this approach because the array is sorted, if it wasn't we would use Math.max() and spread operator. Can you try doing it on that way too?

starter-code/index.js
```diff
@@ -1,10 +1,55 @@
...
+    return this.items[0];
+  }
+                    
+  average() {
+    if (this.length <= 0)
+        return 0;
+                        
+        return this.sum() / this.length;
+  }
+  sum() {
```
sandrabosk: This way is fine too but we learned .reduce(). Try doing it on that way and using fat arrow syntax.

sandrabosk:
Good start of module 2. Keep it this way.

PR #366, opened 2018-10-31
[WEB-PT-MAD2] AnaEstrada

aemartos:
In the max and min method, I tried to use a ternary conditional, but it didn't work, I have been reading, and I saw that is only used to assignment  statements. Is it correct? Or is there any other reasons? I didn't understand it well.
This is what I was trying to do:
the this.length === 0 ? throw new Error("EmptySortedList") : return Math.min(...this.items);

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,5 @@ typings/
 # dotenv environment variables file
 .env
 
+#OSX Sistem information files
+.DS_Store
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,41 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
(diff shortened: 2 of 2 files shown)


ta-web-mad:
Hey Ana! well done 😁🥳

I really like the way you structure the functions, and nice try with the ternary operator!! About that... The ternary operator is a little picky, especially when you want to resolve the if statement with a return and a function. Nevertheless you can do it like this: 

`min() {
    return(
      this.length === 0 ? (()=>{throw new Error("EmptySortedList")})() : Math.min(...this.items)
    )
  }`

**- Error Issue**

The thing is that the ternary operator expect "expressions"... so if you want to resolve the statement sending an error, you have encapsule it into a function expression (a callback) that immediately runs and its return will be the error itself. 

**- Return Issue**

On the other hand, you can't not resolve the ternary operator with a return itself, because the ternary operator evaluates to an expression and expressions can't contain a return statement but you can return the result of the whole ternary operator, in fact this structure is commonly used in react apps ;)

If you have doubts about it, let me know in class and I will explain it ;)
review it by **Cami**

#closed #checked

PR #369, opened 2018-11-01
[WEB-PT-MAD2] AaronReina

AaronReina:


starter-code/index.js
```diff
@@ -1,11 +1,44 @@
...
+    } else {
+      return this.items.reduce((a, b) => a + b, 0) / this.length;
+    }
+  }
+  sum() {
+    if (this.length === 0) {
+      return 0;
+    } else {
+      return this.items.reduce((a, b) => a + b, 0);
+    }
```
ta-web-mad: in this sum() function is not necessary the if statement, you can just return this:
`sum() {
      return this.items.reduce((a, b) => a + b, 0);
  }`

starter-code/index.js
```diff
@@ -1,11 +1,44 @@
...
+      throw new Error("EmptySortedList");
+    } else {
+      return this.items[0];
+    }
+  }
+  average() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.items.reduce((a, b) => a + b, 0) / this.length;
```
ta-web-mad: you can reuse your code like this:
` return this.sum() / this.length;`

because you're doing the same operation like the sum()

ta-web-mad:
Aaron!! Great work!

Well done using the reduce 😁 

review it by Cami #closed #checked

PR #376, opened 2018-11-05
Paris - Corrado

griev04:


ta-web-paris:
You passed all the tests and made use of the new syntax we've learnt. Good job! 👏

PR #377, opened 2018-11-05
Paris - Cécile

Gally999:


starter-code/index.js
```diff
@@ -1,11 +1,52 @@
...
+    }
+  };
+  average() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.sum() / this.length;
+    }
+  }
+  sum() {
```
ta-web-paris: If you wanted to do this as a reduce, you could do it like this  

```
sum () {
  return this.items.reduce(function(sum, el) {
      return sum + el;
    }, 0);
}
```

ta-web-paris:
Good job! 

You passed all the tests and made use of the destructuring that we'd covered in class 💪

PR #378, opened 2018-11-05
[PARIS] Amine

enima2684:


starter-code/index.js
```diff
@@ -1,11 +1,75 @@
...
+    return this._items;
+  }
+
+  set items(value){
+    this._items = value;
+    this.length = this._items.length;
+  }
+
+  add(item) {
+
```
ta-web-paris: you seem to have overcomplicated the add function. 

if you remove this code 
``` 
 if(this.length === 0){
      this.items.push(item);
      this.length++;
      return
    } 
```
all the tests still pass. 

Your code is hard to follow here, a cleaner way to write this function would be 

```
add (items) {
    this.items.push(item);
    this.items = this.items.sort((a, b) => a - b);
    this.length = this.items.length;
}

```

enima2684: 👍 Right !

PR #380, opened 2018-11-05
Paris - Laura

laurastromboni:


starter-code/index.js
```diff
@@ -1,11 +1,55 @@
...
+    })
+    return mySum;
+  }
+
+  average() {
+    var myAverage = 0;
+    if (this.length === 0){
+      throw new Error("EmptySortedList")
+    } else {
+      myAverage = this.sum() / this.length
```
ta-web-paris: here, you could have returned on the same line, you didn't need to create a variable 

``` 
```

starter-code/index.js
```diff
@@ -1,11 +1,55 @@
...
+  min() {
+    if (this.length === 0){
+      throw new Error("EmptySortedList")
+    } else {
+    return this.items[0]
+    }
+  }
+
+  sum() {
+    var mySum = 0;
```
ta-web-paris: You could have used a reduce here 

```
  return this.items.reduce(function(sum, el) {
      return sum + el;
    }, 0); 
```

ta-web-paris:
Your code is good and you passed all the tests!  good job ☀️

PR #381, opened 2018-11-05
Paris - Geoffroy

geoffroybag:


starter-code/index.js
```diff
@@ -1,11 +1,46 @@
...
+  average() {
+    if(this.length>0){
+      return this.sum()/this.length
+    }
+    else {throw new Error("EmptySortedList")
+  }
+  }
 };
 
+var sortedList = new SortedList()
```
ta-web-paris: You don't need this line. You create the variable sortedList but you never use it

starter-code/index.js
```diff
@@ -1,11 +1,46 @@
...
+  }
+  }
+  min() {
+    if(this.length>0){
+      return this.items[0]
+    }
+    else {throw new Error("EmptySortedList")
+  }
+  }
+  sum() {
```
ta-web-paris: You could have used a reduce here. It makes it clearer to people reading your code that you're adding up the values to get a single value. 

```
 return this.items.reduce(function(acc, el) {
      return acc + el;
    }, 0);
```

ta-web-paris:
your code is good! You passed all the tests, good job 👏

PR #385, opened 2018-11-05
Paris - Régis

jourisque:
All Done

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,50 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
-};
+  constructor() {
+    this.items = [];
+    this.length = 0;
+  }
+
+  add(x) {
+    this.items.push(x);
+    this.items.sort( (a,b) => a-b
+    );
+    this.length ++;
+  }
+
+  get(pos) {
+ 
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,50 @@
...
+      throw new Error("EmptySortedList");
+    }
+    return this.items[0];
+  }
+
+  average() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    return (this.items.reduce(  (acc, oneItem) => acc += oneItem
```
ta-web-paris: when calculating the average, you could have called the sum function 

```
 average() {
    if (this.items.length === 0) {
      throw new Error("EmptySortedList");
    }
  }

```

starter-code/index.js
```diff
@@ -1,11 +1,50 @@
...
+    , 0)) / this.items.length;
+  }
+
+  sum() {
+    if (this.items.length === 0) {
+      return 0;
+    }
+    return this.items.reduce(  (acc, oneItem) => acc += oneItem
```
ta-web-paris: when using reduce you don't need += 

```
    return this.items.reduce((acc, oneItem) => acc + oneItem, 0);

```

but good use of reduce and an arrow function!!

ta-web-paris:
you passed all the tests! Good use of the arrow functions that we've learnt in class 🎉

PR #396, opened 2018-11-06
Paris Priyanka Maheshwari

priyaprafful:
done

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,46 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
+  constructor() {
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,46 @@
...
+  min() {
+    if (this.length>0){
+
+    return Math.min(...this.items);
+  }
+  else throw new Error("EmptySortedList");
+}
+  average() {
+    if(this.length>0){
+    let total = this.items.reduce((a,b) =>a+b);
```
ta-web-paris: you've already calculated the sum, so to help keep your code dry you could call it here 

```

```

ta-web-paris:
you've made good use of the arrow functions and destructuring that we've done in class. And your code is easy to follow. good job 🎉

PR #398, opened 2018-11-06
PARIS jeannicolas

blksnk:
initial version, working

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,60 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,60 @@
...
+  		return this.items[0];
+  	};
+  }
+
+  sum() {
+  	if (this.length === 0) {
+  		return 0;
+  	}
+  	else {
+  		let sum = 0;
```
ta-web-paris: to calculate the sum, it's better to use a reduce: 

```
        return acc + curr;
      }, 0);

```

ta-web-paris:
Good job 💪You passed all the tests! I like how you called the sum function when calculating the average

PR #399, opened 2018-11-06
mehdi-louraoui-done

mehdilouraoui:


starter-code/index.js
```diff
@@ -1,11 +1,49 @@
...
+  min() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    return this.items[0];
+  }
+
+  sum() {
+    if (this.items.length === 0) {
+      return 0;
```
ta-web-paris: when doing a reduce you don't need to use +=, you can just use a + 

```
return this.items.reduce((acc, oneItem) => acc + oneItem, 0);
```

starter-code/index.js
```diff
@@ -1,11 +1,49 @@
...
+    }
+    return this.items.reduce((acc, oneItem) => acc += oneItem, 0);
+  }
+
+  average() {
+    if (this.items.length === 0) {
```
ta-web-paris: when calculating the average, you can call the sum function. this will help keep your code dry

```
    return this.sum() / this.items.length;

```

ta-web-paris:
You passed all the tests! Your code is easy to read, and you used reduce well 🙌

PR #400, opened 2018-11-06
Paris - Adèle

AdeleRevert:


starter-code/index.js
```diff
@@ -1,11 +1,60 @@
...
+      var min = Math.min( ...this.items);
+      return min;
+    }
+  }
+
+  average() {
+    if(this.length === 0){
+      throw new Error("EmptySortedList")
+    } else {
+      this.average = this.sum / this.items.length;
```
ta-web-paris: to return the average, you need to call it as a function 

```
return this.sum() / this.items.length
```

starter-code/index.js
```diff
@@ -1,11 +1,60 @@
...
+  add(x) {
+    this.length++;
+    this.items.push(x);
+    this.items.sort(function(a, b) {
+      return a - b;
+    });
+  }
+
+  get(pos) {
+    return this.items[pos];
```
ta-web-paris: here, you need to do 

```
    return this.items[pos - 1];

```

by doing -1 you fix the off by one error. The index starts at 0. So if the array has 10 items in it, the last index is 9.

ta-web-paris:
Your code is good, nice job! I like how you tried to call the sum function in the average function. And you used destructing which is cool 💪

PR #401, opened 2018-11-06
Antoine - Paris

DoneDeal0:
functions seem logic, but the script doesn't pass Mocha test


starter-code/index.js
```diff
@@ -1,11 +1,47 @@
...
+  constructor() {
+    this.items  = [];
+    this.length = 0;
+  }
+  add(x){
+    this.items.push(x);
+    this.length++;
+  }
+
+  get(x){
```
ta-web-paris: for this, you're close. The test wanted
 ```
  get(x) {
    if (this.items.length === 0) {
      throw new Error("OutOfBounds");
    }
    return this.items[x - 1];
  }
```

for the position, you need to -1 because the index of an array starts at 0. Without doing -1, you get an off by one error

File: starter-code/index.js [hunk 2]
```diff
@@ -1,11 +1,47 @@
...
+    if(x > this.length){
+      throw new Error("OutOfBounds")}
+    return this.items[x];
+}
+
+  max(items){
+    if(this.items.length=== 0){
+      throw new Error("EmptySortedList")
+    }
+    return Math.max(...items);
```
ta-web-paris: here you need to  do 
```
```

starter-code/index.js
```diff
@@ -1,11 +1,47 @@
...
+  }
+
+  min(items){
+    if(this.items.length=== 0){
+      throw new Error("EmptySortedList");
+    }
+    return Math.min(...items);
```
ta-web-paris: And here you need to do 

```
    return Math.min(...this.items);

```

starter-code/index.js
```diff
@@ -1,11 +1,47 @@
...
+  }
+
+  sum(items){
+    this.items.reduce(acc, curr)
```
ta-web-paris: you're close with the sum function, but you're missing a few things from your reduce. You never turned it into a function. 

```
  sum() {
      return acc + curr;
    }, 0);
  }
```

starter-code/index.js
```diff
@@ -1,11 +1,47 @@
...
+      return acc + curr;
+    }
+
+
+    average(){
```
ta-web-paris: and you're close here. But you don't call the the sum function. To call the function you need to do 

```
```

In order the pass the other test you just need to add a condition that checks to see if the length is 0

starter-code/index.js
```diff
@@ -1,11 +1,47 @@
...
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
+  constructor() {
+    this.items  = [];
+    this.length = 0;
+  }
+  add(x){
```
ta-web-paris: and in order to pass the  #add(x) and get(i) test, you need to sort the elements. There is a clue in the terminal when you run the test: `should add a second value to SortedList, sorted`. So we just need to do a simple sort and then you'll be able to pass the tests  

```
this.items.sort(function(a, b) {
      return a - b;
    });
```

ta-web-paris:
What you did seemed logically, and you were close, but you were missing a few important details which prevented some of the tests from passing. 

Let us know if you'd like us to review it with us, and if you're unsure of any of the corrections.

PR #404, opened 2018-11-06
Paris Chloe

chloehd:


ta-web-paris:
it's a good start! I Like how you used an arrow function to do the sort :)

PR #499, opened 2019-01-28
Paris: Christophe & Raja

RaproCode:
@grosbegue

It is not easy. It seems we got it right, but not too sure.

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,47 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
-};
+  constructor() {
+    // this.items = listItems;
+    // this.length = listLength;
+    this.items = [];
+    // this.length = 0;
+    this.length = this.items.length;
+  }
+  add(x) {
+    this.items.push(x);
+    this.length++;
+  }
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,47 @@
...
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    return Math.max(items);
+  }
+  min() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    return Math.min(items);
```
ta-web-paris: because we're in a class we need to use `this.items` otherwise items is undefined. 

to pass this test, you can use a spread operator to return the maximum and minimum ```return Math.min(...this.items);```

starter-code/index.js
```diff
@@ -1,11 +1,47 @@
...
+  }
+  average() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+  }
+  sum() {
+    var sumsum = 0;
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
```
ta-web-paris: to pass one more test you need to `return 0`  rather than throwing a new Error. The test message is 'should return 0 for an empty sorted list'

PR #500, opened 2019-01-28
Paris: Thomas, Anoop

brwnrcluse:
finito

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,69 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
-};
+  constructor(items, length) {
+    this.items = [];
+    this.length = this.items.length;
+  }
+
+  add(item) {
+    this.items.push(item);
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,69 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
-};
+  constructor(items, length) {
```
ta-web-paris: you didn't need to pass anything into the constructor, because you never use them

starter-code/index.js
```diff
@@ -1,11 +1,69 @@
...
+
+      return min;
+    }
+  }
+
+  average() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      var total = this.items.reduce(function(sum, current) {
```
ta-web-paris: You could have called the sum() function here, rather than calculating the sum twice

ta-web-paris:
What you've done is good, you passed all of the tests and your code is good 👏

PR #503, opened 2019-01-28
Paris - Aurélie & Estelle <3

Olawinski-e:
Finally we did it ! @aurelieBeno

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,55 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
-};
+  constructor(length, item) {
+    this.length = 0;
+    this.item = [];
+  }
+
+  add(item) {
+    this.item.push(item);
+    this.length += 1;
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,55 @@
...
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.item[0];
+    }
+  }
+
+  sum() {
+    var sum = 0;
+    this.item.forEach(function(oneItem) {
```
ta-web-paris: normally, when calculating a sum, using a reduce is better because it signals better intent. It makes it easier for people to know exactly what you are doing when they first look at your code

starter-code/index.js
```diff
@@ -1,11 +1,55 @@
...
+    });
+    return sum;
+  }
+
+  average() {
+    var average = 0;
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      average = this.sum() / this.length;
```
ta-web-paris: nice! I like how you called the sum() function here

ta-web-paris:
Your code is good and you passed all the tests 😀

PR #506, opened 2019-01-28
Paris - Nadjie & Kevin B

NadjieVal:
@kevbarns


ta-web-paris:
Your code is very good! I like how you made use of calling a function inside a function, spreading an array and arrow function 🎉

PR #510, opened 2019-01-28
Paris - Céline & Armand

abiglary:
@abiglary
@frietpoune


ta-web-paris:
Your code is very good! You passed all the tests and I like how you made use of all the new things we covered in class (spreading, arrow function) and how you called a function within a function 🎉

PR #537, opened 2019-01-29
paris - christina & sam

petitkriket:
@CIBailey


ta-web-paris:
Your code is very good! I like how you made used of arrow functions and spreading an array and how you called a function inside another function 💪

PR #554, opened 2019-01-29
PARIS - WebDev - Kevin M. & Jules F.

F-Jules:
@kevGhub

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,46 @@
 class SortedList {
-  constructor() {}
-  add(item) {}
-  get(pos) {}
-  max() {}
-  min() {}
-  average() {}
-  sum() {}
-};
+  constructor() {
+    this.items = [];
+    this.length = 0;
+  }
+  add(item) {
+    this.items.push(item);
+    this.items.sort();
+    this.length += 1;
+  }
+  get(pos) {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.item[this.length - 1];
+    }
+  }
+  max() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return Math.max(this.items);
+    }
+  }
+  min() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return Math.min(this.items);
+    }
+  }
+  average() {
+    this.items.sum / this.length;
+  }
+  sum() {
+    this.items.reduce(function(acc, curr) {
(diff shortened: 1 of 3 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,46 @@
...
-  average() {}
-  sum() {}
-};
+  constructor() {
+    this.items = [];
+    this.length = 0;
+  }
+  add(item) {
+    this.items.push(item);
+    this.items.sort();
```
ta-web-paris: You're missing something from your sort function which is why some of the other tests are not passing 

```
 this.items.sort(function(a, b) {
      return a - b;
    });
```

starter-code/index.js
```diff
@@ -1,11 +1,46 @@
...
+    this.length += 1;
+  }
+  get(pos) {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.item[this.length - 1];
```
ta-web-paris: you were close here 

```return this.items[pos - 1]; ``` 

this returns the last item of the sorted array

File: starter-code/index.js [hunk 3]
```diff
@@ -1,11 +1,46 @@
...
+    }
+  }
+  max() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return Math.max(this.items);
```
ta-web-paris: to return the max, you can use the spread operator `  return Math.max(...this.items); `  or `this.items[this.length-1]` 

Returning the length -1 works because you've already sorted the array

starter-code/index.js
```diff
@@ -1,11 +1,46 @@
...
+      throw new Error("EmptySortedList");
+    } else {
+      return Math.min(this.items);
+    }
+  }
+  average() {
+    this.items.sum / this.length;
+  }
+  sum() {
+    this.items.reduce(function(acc, curr) {
```
ta-web-paris: you actually need to return the reduce. I know it seems super weird to have two returns 

```
  sum() {
      return acc + curr;
    }, 0);
  }
```

starter-code/index.js
```diff
@@ -1,11 +1,46 @@
...
+  }
+  min() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return Math.min(this.items);
+    }
+  }
+  average() {
+    this.items.sum / this.length;
```
ta-web-paris: I like what you tried to do here, but you want to call the function so you need to add () `this.items.sum() / this.length;`

PR #568, opened 2019-02-11
Monroe Getz FTWD Miami

Mgetz10:


ta-web-mia:
Well done! The 1st test is failing, but I'm pretty sure it's an oversight in the test and not an error in your logic (since all of the others passed fine). Nothing to get too hung up on, good job 👍

PR #703, opened 2019-07-01
MIA-FTWD Taryn Chiang with Billy Wong

tarynchiang:


ta-web-mia:
Nice job! I don't have too much to critique, but let me know if you have any questions. Keep it up 👍 
- Nick

PR #706, opened 2019-07-01
MIA-FTWD-Mitch/Aflredo

mitchmalinin:


MarcosPalaciosIII:
Excellent job on the node intro lab. Node is an important part of javascript and is good to know how it works. It is primarily used to run JS code outside of the browser and build scalable network applications. 

Keep up the excellent work.

PR #718, opened 2019-07-01
MIA FTWD - Taluama Velludo and Andre Cabrera

taluama:


ta-web-mia:
Nice work! Don't worry too much about that one error; there's an issue in the mocha tests that makes it impossible to pass them all unless you edit the test file itself. Keep up the good work 👍
- Nick

PR #720, opened 2019-07-01
MIA-FTWD-mario_salvatierra_modified test_js

mesalvav:


ta-web-mia:
Nice work! And awesome job fixing the test.js so you could pass every test. Keep up the good work, and let me know if you have any questions =]
- Nick

PR #731, opened 2019-07-03
FTWD-Miami-Serena Dominguez

SDominguez21:


ta-web-mia:
Be careful about calling a function within itself; functions can be called recursively like this, but in this case it's throwing an error:
image

 You're on the right track and passing 5 of the tests, so only go back to it if you have extra time. Keep it up, and let me know if you have any questions =]
- Nick

PR #890, opened 2019-11-05
[PAR_WDFT] Nicolas

Empty-n:
Done :)

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,62 @@
 class SortedList {
-    constructor() {}
-    add(item) {}
-    get(pos) {}
-    max() {}
-    min() {}
-    average() {}
-    sum() {}
-  };
-  
-  module.exports = SortedList;
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
+  add(item) {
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,62 @@
...
+      }
+    }
+    return result / this.items.length;
+  }
+  sum() {
+    var result = 0;
+    if (this.items.length == 0) {
+      return [];
+    } else {
+      for (i = 0; i < this.items.length; i++) {
```
ta-web-paris: you forget to define i ^^ for (let i=0)
so the test is not working

starter-code/index.js
```diff
@@ -1,11 +1,62 @@
...
+        return a - b;
+      });
+      return this.items[0];
+    }
+  }
+  average() {
+    var result = 0;
+    if (this.items.length == 0) {
+      throw new Error("EmptySortedList");
+    } else {
```
ta-web-paris: you can reuse your sum function instead of rewrite it

ta-web-paris:
almost perfect ;)

PR #893, opened 2019-11-05
[PAR WDFT Evgeny]

ESR93:
Done =)

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,55 @@
 class SortedList {
-    constructor() {}
-    add(item) {}
-    get(pos) {}
-    max() {}
-    min() {}
-    average() {}
-    sum() {}
+    constructor() {
+      this.items=[]
+      this.length=0
+    }
+    add(item) {
+      this.items.push(item)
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,55 @@
...
+      this.length ++
+    }
+    get(pos) {
+      return this.items[pos]
+    }
+    max() {
+      let mx=0
+      if(this.length!==0)
+      {
+        this.items.forEach(element => {
```
ta-web-paris: you can directly use array.sort and then select the index 0 
(same for min)

starter-code/index.js
```diff
@@ -1,11 +1,55 @@
...
+            mn=element
+        });
+        return mn
+      }
+      else throw new Error("EmptySortedList")
+    }
+    average() {
+      if(this.length!==0)
+      {
+        let sum=0
```
ta-web-paris: instead of redoing the sum you can directly call your sum function

ta-web-paris:
Good Job!

PR #894, opened 2019-11-05
update tests, solve OutOfBounds issue, turn to es6

sandrabosk:
@mjarraya Please check the updated tests. 
Find here the proposed solution: https://gist.github.com/sandrabosk/1dc434be87b4efc03238e53ffae1376f

### DIFF
--- a/starter-code/test/test.js
+++ b/starter-code/test/test.js
@@ -1,6 +1,6 @@
 // Require the file which will be tested
 const SortedList = require('../index.js');
-// Load the Chai Assertion Library
+// Load the Assertion Library
 const assert = require('assert');
 
 describe('SortedList', () => {
@@ -125,7 +125,7 @@ describe('SortedList', () => {
       sl = new SortedList();
     });
 
-    it('should return sum of all elements in the array', () => {
+    it('should return the sum of all elements in the array', () => {
       sl.add(1);
       sl.add(2);
       sl.add(3);
--- a/starter-code/test/test.js
+++ b/starter-code/test/test.js
@@ -55,28 +55,26 @@ describe('SortedList', () => {
    
   });
 
-
-  // I really don't understand why we have to specify this?
-  // describe('#add(x) and get(i)', function() {
-  //   var sl;
-  //   beforeEach(function(){
-  //     sl = new SortedList();
-  //   });
-  //   it('should add a second value to SortedList, sorted', function() {
-  //     sl.add(20);
-  //     sl.add(10);
-  //     assert.equal(sl.get(0), 10);
-  //     assert.equal(sl.get(1), 20);
-  //   });
-  //   it('should add a third value to SortedList, sorted', function() {
-  //     sl.add(30);
-  //     sl.add(20);
-  //     sl.add(10);
-  //     assert.equal(sl.get(0), 10);
-  //     assert.equal(sl.get(1), 20);
-  //     assert.equal(sl.get(2), 30);
-  //   });
-  // });
+  describe('#add(x) and get(i)', () =>  {
+    let sl;
+    beforeEach(() => {
+      sl = new SortedList();
+    });
+    it('should add a second value to SortedList, sorted',() => {
+      sl.add(20);
+      sl.add(10);
+      assert.equal(sl.get(0), 10);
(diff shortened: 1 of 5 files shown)


starter-code/test/test.js
```diff
@@ -1,73 +1,72 @@
...
       assert.equal(sl instanceof SortedList, true);
+      assert.equal(sl.length, 0);
+      assert(Array.isArray(sl.items))
     });
   });
 
-  describe('#add(x)', function() {
-    var sl;
-    beforeEach(function(){
+  describe('#add(x)', () => {
```
mjarraya: This unit should also test the spec that the list of items is sorted

starter-code/test/test.js
```diff
@@ -1,73 +1,72 @@
...
+      }, Error)
     });
+
+   
   });
 
-  describe('#add(x) and get(i)', function() {
-    var sl;
-    beforeEach(function(){
+  describe('#add(x) and get(i)', () =>  {
```
mjarraya: these specs could be integrated in the unit that describes `add`

starter-code/test/test.js
```diff
@@ -1,73 +1,72 @@
...
     });
 
-    it('should return the element in that position', function() {
-      var foo = 10;
-      for(let i=0; i<200; i++) {
-        sl.add(foo*i);
-        assert.equal(sl.get(i), foo*i);
-      }
+    it('should return an OutOfBounds exception if there is no element in that position', () => {
+      assert.throws(() => {
```
mjarraya: `assert.throws(() => { sl.get(4); }, Error, "OutOfBounds");`

It's not relevant to create a condition for equality of the return value of the method with the string `"undefined"` -which should never be reached-. `assert.throws()` only needs the function to be called with the right parameters, and we can pass in a third parameter to `.throws()` the error's name

starter-code/test/test.js
```diff
@@ -77,97 +76,90 @@ describe('SortedList', function() {
...
     });
-  });
 
-  describe('#min()', function() {
-    var sl;
-    beforeEach(function() {
-      sl = new SortedList();
+    it('should return an EmptySortedList exception if there is no elements in the array', () => {
+       assert.throws(() => {
+        if (sl.max([]) === 'undefined') {
```
mjarraya: `assert.throws(() => { sl.max([])) }, Error, "EmptySortedList");`

starter-code/test/test.js
```diff
@@ -77,97 +76,90 @@ describe('SortedList', function() {
...
+    it('should return the min (lowest) value in the array', () => {
       sl.add(10);
       sl.add(20);
-
       assert.equal(sl.min(), 10);
     });
+
+    it('should return an EmptySortedList exception if there is no elements in the array', () => {
+      assert.throws(() => {
+        if (sl.min([]) === 'undefined') {
```
mjarraya: see above

starter-code/test/test.js
```diff
@@ -77,97 +76,90 @@ describe('SortedList', function() {
...
     })
 
-    it('should return the average of elements in the array', function() {
-      for(i=0; i<101; i++){
-        sl.add(i*2);
-      }
-      assert.equal(sl.average(), 100);
+    it('should return an EmptySortedList exception if there are no elements', () => {
+      assert.throws(() => {
+        if (sl.avg([]) === 'undefined') {
```
mjarraya: see above

starter-code/test/test.js
sandrabosk: This test requires using `add()` and `get()` but in the moment we define `add()`, we still don't have `get()` defined. That's why I kept it after. Maybe I am not thinking in the right direction. Please let me know if I am wrong. Thank you.

sandrabosk:
- updated tests that required a change
- removed global mocha installation (it's already using 6.x.x)

- **pending**: should additional test on _add ()_ be merged with _add()_?

starter-code/test/test.js
mjarraya: The students implement `add()` without the sorting, and the tests pass. Then they don't necessarily see what was wrong with their initial implementation (does it come from the `add()` ? from the `get()` ?) and have to backtrack to implement the `add` method correctly.

PR #898, opened 2019-11-06
Par WDFT Noemie

NoemieNT:
missing last 2 iterations

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,11 +1,29 @@
 class SortedList {
-    constructor() {}
-    add(item) {}
-    get(pos) {}
-    max() {}
-    min() {}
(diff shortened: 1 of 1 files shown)


starter-code/index.js
```diff
@@ -1,11 +1,29 @@
...
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else return Math.max(...this.items);
+  }
+  min() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else return Math.min(...this.items);
+  }
+  sum() {}
```
ta-web-paris: remember reduce? Try to use it for sum()

and then call back sum() into average() ;)

PR #943, opened 2019-11-20
update tests and readme

sandrabosk:
Hi @josecarneiro, as I mentioned earlier together with Montasar I made some changes in this lab. Now I have reviewed your PR and took some stuff that we missed to update so this PR is the result of our mutual work. When you find a moment please review it.
Here is solution that might help you test it: https://gist.github.com/sandrabosk/1dc434be87b4efc03238e53ffae1376f

**Changed files:**
- README.md
- test/index.spec.js

Thank you!


josecarneiro:
Hi @sandrabosk and @mjarraya,

Awesome work! You really gave this lab the overhaul it needed. If we keep running this pace, we won't have much maintenance work to do in the months ahead.

I made an additional commit with a few tiny fixes in the `README.md` that I had previously overlooked in my own branch.

I also added an additional check to the sorting test case that I only realized was necessary when asked about it by a student: When sorting the numbers `[10, 20, 30]`, just using `this.items.sort()` would make the test pass, although that's not an acceptable solution, since sorting the numbers `[10, 20, 30, 100]` without passing a sorting function to the `sort` method would return `[10, 100, 20, 30]`. As such, I added two lines of code to the test to ensure that the sorting is performed as desired.

I am merging the pull request to master and closing it.

Cheers!

PR #946, opened 2019-12-02
MIA OCT 2019 Johnny Nguyen

jnguyen3436:


ta-web-mia:
Nice work, and very cool avg()! Being able to run JavaScript in your terminal using node can be incredibly useful for debugging. Plus it means you won't have to keep a repl.it tabbed at all times :P

Keep up the great work!

- Nick

PR #947, opened 2019-12-02
FTWD MIA - October 2019 - Pradeepa Shanmugasundaram

pradeepa24:


ta-web-mia:
Nice work, all test cases passed! Being able to run JavaScript in your terminal using node can be incredibly useful for debugging. Plus it means you won't have to keep a repl.it tabbed at all times :P

Keep up the great work!

- Nick

PR #955, opened 2019-12-02
MIA-ftwd-oct2019 Gil (gmsy)

gmsy:


ta-web-mia:
Nice work, I really like how much you were able to abstract your code within the functions! Being able to run JavaScript in your terminal using node can be incredibly useful for debugging. Plus it means you won't have to keep a repl.it tabbed at all times :P

Keep up the great work!

- Nick

PR #1013, opened 2020-01-28
[BER-FT] Elsa

ElsaFil:


starter-code/index.js
```diff
@@ -1,17 +1,46 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = 0;
```
DalinaLeila: `this.length = this.items.length`

starter-code/index.js
```diff
@@ -1,17 +1,46 @@
...
+  }
 
-  add(item) {}
+  add(item) {
+    this.length++;
```
DalinaLeila: it works if you only push one item every time (in this case it will work). Maybe better to write `this.length = this.items.length` after you pushed x into the array and sorted it

starter-code/index.js
```diff
@@ -1,17 +1,46 @@
...
+    this.items.push(item);
+    this.items.sort((a, b) => a - b);
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos >= this.items.length) {
```
DalinaLeila: maybe also check if its below 0, because it would return undefined.

starter-code/index.js
```diff
@@ -1,17 +1,46 @@
...
+    }
+    return this.items[pos];
+  }
 
-  max() {}
+  max() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    return this.items[this.items.length - 1];
```

starter-code/index.js
```diff
@@ -1,17 +1,46 @@
...
+  }
 
-  min() {}
+  min() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    return this.items[0];
```

starter-code/index.js
```diff
@@ -1,17 +1,46 @@
...
-  get(pos) {}
+  get(pos) {
+    if (pos >= this.items.length) {
+      throw new Error("OutOfBounds");
+    }
+    return this.items[pos];
+  }
 
-  max() {}
+  max() {
```
DalinaLeila: I think for max and min you should return the highest/lowest number in the array. right now this works because the array is sorted but maybe safer to use Math.max() Math.min() :)

starter-code/index.js
ElsaFil: ahhh true!

PR #1014, opened 2020-01-28
[BER FT 01-2020] Konstantin

kdauer:


starter-code/index.js
```diff
@@ -1,17 +1,54 @@
...
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort((a, b) => a - b);
+    this.length = this.items.length;
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (this.length < pos) {
```
DalinaLeila: maybe also check if pos is less than 0 to also check for negative numbers.

starter-code/index.js
```diff
@@ -1,17 +1,54 @@
...
+    } else {
+      return Math.min(...this.items);
+    }
+  }
 
-  sum() {}
+  sum() {
+    if (this.length === 0) {
+      return 0;
+    } else {
```
DalinaLeila: no need for the else because you have a return

PR #1016, opened 2020-01-28
done

Norizinus:


starter-code/index.js
```diff
@@ -1,17 +1,48 @@
...
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort((a, b) => a - b);
+    this.length = this.items.length;
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos > this.length - 1) {
```
DalinaLeila: Nice! It would also be good the check if the position is bigger than 0, otherwise, it will return undefined.

starter-code/index.js
```diff
@@ -1,17 +1,48 @@
...
+      throw new Error("OutOfBounds");
+    }
+    return this.items[pos];
+  }
 
-  max() {}
+  max() {
```
DalinaLeila: for max and min: it works to select the last and first element but maybe it is better to use Math.max and Math.min here in order to also pick the highest/lowest number if the array is not sorted.

starter-code/index.js
```diff
@@ -1,17 +1,48 @@
...
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    }
 
-  avg() {}
+    return this.items[0];
+  }
+
+  sum() {
+    return this.items.reduce((accumulator, value) => value + accumulator, 0);
```

PR #1017, opened 2020-01-28
[BER-FT-012020] Mia L.

mialav:


starter-code/index.js
```diff
@@ -1,17 +1,53 @@
 class SortedList {
-  constructor() {}
+  constructor(...items) {
```
DalinaLeila: no need to pass something here

starter-code/index.js
```diff
@@ -1,17 +1,53 @@
...
+    this.items = items;
```
DalinaLeila: `this.items = []; 
this.length = this.items.length`

starter-code/index.js
```diff
@@ -1,17 +1,53 @@
...
+    this.length = items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort(function(a, b){return a - b});
```
DalinaLeila: use ES6 :) try to get familiar with the arrow function 

`    this.items.sort((a, b) => {return a - b});
`

starter-code/index.js
```diff
@@ -1,17 +1,53 @@
...
+    this.length = this.items.length;
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos < this.length) {
```
DalinaLeila: also check if pos is below 0, otherwise it would return undefined in this case.

starter-code/index.js
```diff
@@ -1,17 +1,53 @@
...
+    } else {
+      throw new Error("OutOfBounds");
+    }
+    
+  }
 
-  max() {}
+  max() {
+    if (this.length > 0) {
+      return Math.max(...this.items);
```

starter-code/index.js
```diff
@@ -1,17 +1,53 @@
...
+    } else {
+      throw new Error("EmptySortedList");
+    }
+  }
 
-  min() {}
+  min() {
+    if (this.length > 0) {
+      return Math.min(...this.items);
```

starter-code/index.js
```diff
@@ -1,17 +1,53 @@
...
+    } else {
+      throw new Error("EmptySortedList");
+    }
+  }
 
-  sum() {}
+  sum() {
+      return this.items.reduce((acc, val) => {
```
DalinaLeila: if you will not assign the reduce to a variable, you can also only return at the beginning and it should still give you the correct sum

PR #1021, opened 2020-01-28
[BER-FT] - Marcel

Marsu-R:


starter-code/index.js
```diff
@@ -1,17 +1,57 @@
...
+     return this.items.reduce((acc, val) => acc + val)
+    }
+  }
 
-  avg() {}
+  avg() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return (this.items.reduce((acc, val) => acc + val)) / this.items.length;
```
fonsogms: you could have used sum() here for reusal instead of writing the same code again.
Still nice job! 
Keep it up!

PR #1022, opened 2020-01-28
[BER-FT-012020]) - Victoria B.

VictoriaBlat:


starter-code/index.js
```diff
@@ -1,17 +1,56 @@
...
+      return this.items.reduce((a, b) => a + b, 0);
+    }
+  }
 
-  avg() {}
+  avg() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.items.reduce((a, b) => a + b, 0) / this.length;
```
fonsogms: you could have reuse the sum() result here for cleaner code.
But after that the lab is well done.
Keep up the good work or the TAs will cry

PR #1024, opened 2020-01-28
[BER-FT] - Tim // Done

timbotimber:
Node is Life

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,17 +1,54 @@
 class SortedList {
-  constructor() {}
+  constructor(items, length) {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.length = this.items.length;
+    this.items.sort((a, b) => a - b);
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (this.length < pos) {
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,17 +1,54 @@
...
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.length = this.items.length;
+    this.items.sort((a, b) => a - b);
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (this.length < pos) {
```
fenrew: Check if the position is less than 0 as well as this would return undefined

starter-code/index.js
```diff
@@ -1,17 +1,54 @@
...
+      throw new Error("OutOfBounds");
+    } else {
```

starter-code/index.js
```diff
@@ -1,17 +1,54 @@
...
+      return this.items.reduce((accu, value) => accu + value);
+    }
+  }
 
-  avg() {}
+  avg() {
+    if (this.length <= 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.items.reduce((accu, value) => accu + value) / this.length;
```
fenrew: Keep in mind that you can use the function: this.sum() so you don't have to reduce twice

PR #1025, opened 2020-01-28
#wd-ft-ber-01-20 martin done

martinschmal:


starter-code/index.js
```diff
@@ -1,17 +1,66 @@
...
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.length = this.items.length;
+    this.items.sort(function(a, b) {
```
fenrew: Nicer if you use ES6 here as we learned yesterday, but this still works

starter-code/index.js
```diff
@@ -1,17 +1,66 @@
...
+      return a - b;
+    });
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos > this.items.length) {
```
fenrew: Check if the position is less than 0 as well as this would return undefined

starter-code/index.js
```diff
@@ -1,17 +1,66 @@
...
+      throw new Error("OutOfBounds");
+    } else return this.items[pos];
```

starter-code/index.js
```diff
@@ -1,17 +1,66 @@
...
+  }
 
-  max() {}
+  max() {
+    if (this.items.length == 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      this.items.sort(function(a, b) {
```
fenrew: I believe the this.items list is sorted every time  you add a new element to it. You will not have to sort it again here

starter-code/index.js
```diff
@@ -1,17 +1,66 @@
...
-  avg() {}
+      return this.items[this.items.length - 1];
+    }
+  }
+
+  sum() {
+    if (this.items.length == 0) {
+      return 0;
+    } else {
+      let result = this.items.reduce((a, b) => a + b, 0);
```
fenrew: If you don't have any default value with reduce, then it is automatically set to 0

starter-code/index.js
```diff
@@ -1,17 +1,66 @@
...
+
+      return result;
+    }
+  }
+
+  avg() {
+    if (this.items.length == 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      let result = this.items.reduce((a, b) => a + b, 0) / this.items.length;
```
fenrew: Keep in mind that you can use the function: this.sum() so you don't have to reduce twice

PR #1026, opened 2020-01-28
[BER-FT] Jason

jfrank325:


starter-code/index.js
```diff
@@ -1,17 +1,57 @@
...
-  add(item) {}
+  add(x) {
+    this.items.push(x);
+    this.items.sort((a, b) => a - b);
+    this.length = this.items.length;
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos >= this.length || pos < 0) {
```
fenrew: Nice, you included the less than 0!

starter-code/index.js
```diff
@@ -1,17 +1,57 @@
...
+      throw new Error('OutOfBounds');
+    } else {
```

PR #1027, opened 2020-01-28
done

shaulitto:
Shaul Web Dev january berlin 2020

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,17 +1,51 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort(function(a, b) {
+      return a - b;
+    });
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,17 +1,51 @@
...
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort(function(a, b) {
```
fenrew: Nicer if you use ES6 here as we learned yesterday, but this still works

starter-code/index.js
```diff
@@ -1,17 +1,51 @@
...
+    }
+    return Math.min(...this.items);
+  }
 
-  sum() {}
+  sum() {
+    if (this.items.length === 0) {
+      return 0;
+    }
+    return this.items.reduce((a, b) => a + b, 0);
```
fenrew: Remember: When using reduce and dont set a default value, it'll be automatically be set to 0.

PR #1028, opened 2020-01-28
[BER - FT] Laurel Wright

lwrocks:
Done - will continue to work on the ones I was stuck on.

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,17 +1,36 @@
 class SortedList {
-  constructor() {}
+  constructor(items, length) {
+    this.items = [];
+    this.length = 0;
+  }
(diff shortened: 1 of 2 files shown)


starter-code/index.js
```diff
@@ -1,17 +1,36 @@
...
+  }
 
-  max() {}
+  max() {
+    throw new Error("EmptySortedList"); 
+  }
 
-  min() {}
+  min() {
+    return this.items.sort((a, b) => (a - b))[0];
```
itstheandre: here you should have used the Math.min
```js
return Math.min(...this.items)
```

It still needs to check for the bad cases, though

PR #1029, opened 2020-01-28
[BER - FT] Stjepan

stifamajstor:


starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+    this.items = this.items.sort((a, b) => a - b);
+  }
 
-  get(pos) {}
+  get(pos) {
+    let value = this.items[pos];
+    if (pos === this.items[pos]) {
+      return value;
+    } else {
+      throw new Error("OutOfBounds");
```
itstheandre: This function is a little confusing. You could have checked directly for the pos value and from there do your returns.

It's totally correct, but a little confusing for my head, at least

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+    }
+  }
 
-  max() {}
+  max() {
+    let max = this.items.reduce((a, b) => {
```
itstheandre: you used the Math.max. No need to iterate using the reduce.
```js
```
Let the javascript magic do the rest

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+    } else {
+      throw new Error("EmptySortedList");
+    }
+  }
 
-  min() {}
+  min() {
+    let min = this.items.reduce((a, b) => {
+      return Math.min(a, b);
+    });
```
itstheandre: Same as above

PR #1030, opened 2020-01-28
[BERFT0120]-Vaso

vasilikibalidou:


starter-code/index.js
```diff
@@ -1,17 +1,50 @@
...
+      throw new Error("OutOfBounds");
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    return this.items[this.length - 1];
```
itstheandre: Even though this is correct, you sould have tried using the Math.max
```js
````
Its correct, but a little hacky 😆

PR #1031, opened 2020-01-28
BE_FT_012020_Akua

akups:


starter-code/index.js
```diff
@@ -1,17 +1,57 @@
...
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort((a, b) => {
+      return a - b;
```
itstheandre: Here you could have used your new found ES6 skills :)
```js
this.items.sort((a, b) => a -b )
```

PR #1032, opened 2020-01-28
[BER, FTWD, 012020] Mohamed Ciwan

ciwan-aram:


starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+    }
+    return item;
+  }
 
-  max() {}
+  max() {
+    if (this.items == 0) {
+      throw new Error('EmptySortedList');
+    }
+    return this.items[this.length - 1];
```
itstheandre: Even though this works. Ideally you should have used Math.max
```js
```

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+    }
+    return this.items[0];
+  }
 
-  sum() {}
+  sum() {
+    let total = 0;
+    for (let i = 0; i < this.length; i++) {
+      total += this.items[i];
+    }
```
itstheandre: Even though this is correct, the reduce would have been a cleaner approach :)

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
-  avg() {}
+  avg() {
+    let total = 0;
+    if (this.items == 0) {
+      throw new Error('EmptySortedList');
+    }
+    for (let i = 0; i < this.length; i++) {
+      total += this.items[i];
+    }
+    return total / this.length;
```
itstheandre: You could have DRY(Don't repeat yourself)'ed this.

and just:

```js
return this.sum()/ this.items.length
```

ciwan-aram: I tried this but somehow couldn't make it work! Something was missing and I didn't have the time to search. 
Thank you for all your comments :)

PR #1034, opened 2020-01-28
[BER-FT-012020]) - Saeka 

saekaisozumi:


starter-code/index.js
```diff
@@ -1,17 +1,55 @@
 class SortedList {
-  constructor() {}
+  constructor(items, length) {
+    this.items = [];
+    this.length = 0;
```
DalinaLeila: `this.length = this.items.length`

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+  }
 
-  add(item) {}
+  add(item) {
+    this.items[this.length] = item;
+    this.items.sort((a, b) => a - b);
+    this.length++;
```
DalinaLeila: this works, for this lab but keep in mind if you would like to add more than one item to your array and will just increment the length it will not be the correct length. 

you would need to write:
`this.length = this.items.length`

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos <= this.length) {
```
DalinaLeila: also check if the pos is below 0, negative values are not working to be used as an index so it would return undefined.

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+      throw new Error("OutOfBounds");
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return Math.max(...this.items);
```

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+    }
+  }
 
-  min() {}
+  min() {
+    if (this.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return Math.min(...this.items);
```

starter-code/index.js
```diff
@@ -1,17 +1,55 @@
...
+    }
+  }
 
-  sum() {}
+  sum() {
+    if (this.length === 0) {
```
DalinaLeila: no need to have the first if. Even if the array has 0 items, the reduce method will then simply return 0.

PR #1079, opened 2020-02-11
[BCN-PTWD-122019] - Thibault Delfaud

Thibault-d:
All exercises completed ; only 1 error remainig that I don't really understand:

> 

  1) SortedList
       #add(x)
         should add a single value to SortedList:

      AssertionError [ERR_ASSERTION]: 0 == 1
      + expected - actual

      -0
      +1

      at Context.<anonymous> (test\index.spec.js:25:14)
      at processImmediate (internal/timers.js:439:21)

### DIFF
--- a/starter-code/index.js
+++ b/starter-code/index.js
@@ -1,7 +1,15 @@
 class SortedList {
-  constructor() {}
+  constructor(items, length) {
+    this.items = [];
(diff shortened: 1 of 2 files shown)


AlexBHdez:
If you add an item, you have to increase the items length property.

PR #1098, opened 2020-02-29
Done

danielhoogerwerf:


starter-code/index.js
```diff
@@ -1,17 +1,56 @@
...
+    }
+  }
 
-  avg() {}
+  avg() {
+    if (this.items.length <= 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      const reducer = (acc, value) => acc + value;
+      return this.items.reduce(reducer) / this.items.length;
```
LloydC: You can also use "return this.sum()/this.items.length;"

PR #1100, opened 2020-02-29
Done! [AMS-PT-012020] - Alex

alexpanko:


starter-code/index.js
```diff
@@ -1,17 +1,66 @@
...
+          return a + b
+        }, 0);
+  }
 
-  avg() {}
+  avg() {
+    if (this.items.length == 0 || this.items.length == undefined) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.items.reduce(function(a,b){
```
LloydC: You can also use "return this.sum()/this.length;"

PR #1101, opened 2020-02-29
Anastasiia Kozlova PT Amsterdam

anastezia-kz:


starter-code/index.js
```diff
@@ -1,17 +1,54 @@
...
+      return this.items.reduce(((total, curr)  => total += curr),0)
+    }
+  }
 
-  avg() {}
+  avg() {
+    if(this.length === 0){
+      throw new Error("EmptySortedList");
+  }else{
+    return this.items.reduce(((total, curr)  => total += curr),0)/this.length
```
LloydC: You can also use "return this.sum()/this.length;"

PR #1102, opened 2020-02-29
done

fhoman:


starter-code/index.js
```diff
@@ -1,17 +1,42 @@
...
+    return Math.min.apply(null, this.items)
+  }
+  sum() {
+    if (this.items.length === 0) {
+      return 0;
+    }
+    return this.items.reduce((accumulator, currentValue) => accumulator + currentValue)
+  }
+  avg() {
+    return this.items.reduce((accumulator, currentValue) => accumulator + currentValue) / this.items.length
```
LloydC: You can also use "return this.sum()/this.items.length;"

PR #1149, opened 2020-03-29
WDPT - INTRO NODE - Olga K - done

Olga-OK:


starter-code/index.js
```diff
@@ -1,17 +1,52 @@
...
+
+  sum() {
+    let sum = 0
+    for (let i = 0; i < this.items.length; i ++){
+      sum += this.items[i]
+    }
+  }
 
   avg() {}
 }
```
numbing: so far really good, nice job just avg is missing

PR #1205, opened 2020-04-14
[PAR-WDTF-0320] – Rebecca Norén 🐯 ongoing

rebeccanoren:
Done! 👯


TatijanaRajcic:
Hey Rebecca! Thanks for your lab! It's really good, I don't have much to add!
You don't need that specific condition in your sum method: `if (!this.length) return 0;`
You put a default value in your reduce (0), so it's gonna return 0 if the array is empty.
Great work !

PR #1299, opened 2020-07-23
[REMOTE-WDPT-0602] - Michel Ribbens

AmanoLX:


index.js
```diff
@@ -1,17 +1,57 @@
...
+
+    return sum; // return the sum of the array
+  }
+
+  avg() {
+    if(this.length === 0) { // if array is empty
+      throw new Error('EmptySortedList'); // throw error message
+    }
+
+    return this.sum() / this.length;
```
valmgisbert, line 53: good on reusing functions!

index.js
```diff
@@ -1,17 +1,57 @@
...
+  min() {
+    if(this.length === 0) { // if array is empty
+      throw new Error('EmptySortedList'); // throw error message
+    }
+    return this.items[0]; // array is zero-based so first item has position 0
+  }
 
-  avg() {}
+  sum() {
+    if (this.length === 0) { // if array is empty
```
valmgisbert, line 37: your reduce method already takes into account any eventuality of the length, so this is like a step above what you need.

valmgisbert:
Great job, Michel! I would say only to try and stay consistent in the position of your error handlings along your file; taking into account in which position you feel like it would be easier to read (before the method to know the ways it could go wrong, or after to first be aware of what your function does and then think of the errors); up to you!

PR #1300, opened 2020-07-23
[REMOTE-WDPT-0602] - Sunitha Mannam

SunithaMannam:


index.js
```diff
@@ -1,17 +1,78 @@
...
+  //Iteration 1: constructor()
+  constructor(items, length) {
+       this.items = [];
+      this.length = 0;     
+  }
 
-  get(pos) {}
+  // Iteration 2: add(item)
+  add(item) {
+    if (typeof item != 'undefined') {
```
itstheandre, line 11: for this method, there was no usecase where an item would be falsy.
But if you want to prepare for this, instead of wrapping the usable / correct code inside a big if statement, you can prepare beforehand like this:
```js
if (!item) //this handles any falsy values,
``` 
and you throw an error there, and then do your logic then

index.js
```diff
@@ -1,17 +1,78 @@
...
+    }
+  }
 
-  sum() {}
+  //Iteration 4: max()
+  max() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    } else {
+      return this.items.sort((a, b) => (a - b)).slice(this.length - 1);
```
itstheandre, line 42: you shouldnt have to sort it, becasuse every time you add something, you immediately are sorting them, so no need to re-do that method

index.js
```diff
@@ -1,17 +1,78 @@
...
+  }
+
+  sum() {
+    if (this.items.length === 0) {
+      return 0;
+    } else {
+      return this.items.reduce((sum, item) => {
+        return sum + item;
+      }, 0);
+    }
```
itstheandre, line 62: You dont need to do the first if statement.
If you loop with reduce, over an empty array, the final value will be its accumulator (either the emtpy array itself) or the initial value you give in the reduce:
```js
return this.items.reduce((acc, val) => {
    acc + val
}, 0)
```
in this case, the initial value would be 0.
Less code all around, but still correct

PR #1301, opened 2020-07-23
[REMOTE-WDPT-0620] - Debora Crescenzo

NubivagantYansa:
done

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,37 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    (this.items = []), (this.length = this.items.length);
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort((a, b) => a - b);
+    this.length = this.items.length;
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos > this.items.length) throw new Error("OutOfBounds");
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,17 +1,37 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    (this.items = []), (this.length = this.items.length);
```
itstheandre, line 3: I have literally, never seen this before. But if it works, thank you for showing me something new.
🤷‍♂️

index.js
```diff
@@ -1,17 +1,37 @@
...
 
-  sum() {}
+  sum() {
+    return this.items.reduce((acc, value) => (acc += value), 0);
+  }
 
-  avg() {}
+  avg() {
+    if (!this.items.length) throw new Error("EmptySortedList");
+    return this.items.reduce((acc, value) => (acc += value), 0) / this.length;
```
itstheandre, line 33: The only thing here would have been to reuse the sum method. no need to reinvent the wheel, or in this case, rewrite the sum method.
```js
```

index.js
NubivagantYansa: That was auto-formatted by Pritter. It passed the tests and I thought to leave it there :)

PR #1302, opened 2020-07-24
[REMOTE-WDPT-0602] - Rodrigo Hernández

Rodrigohernandezsan:
Lab Done!

Thanks

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,46 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,17 +1,46 @@
...
+
+  min() {
+    if (this.length === 0) throw new Error('EmptySortedList');
+    return this.items[0];
+  }
+
+  sum() {
+
+    if (this.length === 0) return 0;
+    return this.items.reduce((acc, curVal) => acc + curVal);
```
itstheandre, line 34: one quick thing, you can try to loop over an empty array. 
If you reduce like this:
```js
return this.items.reduce((acc,curVal) => acc + curVal, 0)
```
Even if the array is empty, the reduce will default to its "initial value", in this case, the 0

PR #1303, opened 2020-07-24
[REMOTE-WDPT-0602] - Alejandro González

alejanglez:
Everything done and passed ! thank you!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,50 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,17 +1,50 @@
...
+  sum() {
+    return this.items.reduce((acc, value) => (acc += value), 0);
+  }
 
-  avg() {}
+  avg() {
+    if (this.items.length === 0){
+      throw new Error('EmptySortedList');
+    } else {
+      return this.items.reduce((acc, value) => (acc += value), 0) / this.items.length;
```
itstheandre, line 45: Damn, Alejandro.
You were so close to perfection!
Im kidding, the only feedback is really that you could have reused the sum method and use this.length
```js
```

PR #1304, opened 2020-07-24
[REMOTE-WDPT-0602] - Matthew Howard

mattalanhoward:


index.js
```diff
@@ -1,17 +1,54 @@
...
-  add(item) {}
+  add(item) {
+    this.items.push(item)
+    this.items.sort(function(a, b){return a-b})
+    this.length ++
+  }
 
-  get(pos) {}
+  get(pos) {
+    if(pos > this.length-1){
```
itstheandre, line 14: less key strokes would be:
```js
if (pos>= this.length)
```
But also works :)

PR #1306, opened 2020-07-24
[REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:


index.js
```diff
@@ -1,17 +1,46 @@
...
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort((a,b) => a-b);
+    this.length++;
```
itstheandre, line 10: in this case, even though its correct, I would rather reassign the value, instead of directly changing it.

```js
this.length = this.items.length
```

Might be personal preference I guess

index.js
```diff
@@ -1,17 +1,46 @@
...
+  sum() {
+    return this.items.reduce((acc, item) => acc + item, 0);
+  }
 
-  avg() {}
+  avg() {
+    if (this.length === 0) {
+      throw new Error('EmptySortedList');
+    }
+    return this.items.reduce((acc, item) => acc + item, 0) / this.length;
```
itstheandre, line 42: faster approach here would be to reuse the sum method you had already written. No need to repeat code.

```js
```

PR #1310, opened 2020-07-24
[WDPT REMOTE 0602] Andre Bacalhau

AndrewMastercode:


index.js
```diff
@@ -1,17 +1,60 @@
...
+      return this.items[pos]
+    } else {
+      throw new Error('OutOfBounds')
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.length>0) {
+      var max = this.items.reduce(function(a, b) {
```
valmgisbert, line 23: In this function, you are doing the same thing twice: your reduce is comparing and grabbing the max value, and you also return the max value with the Math method; you should stick to one only.

index.js
```diff
@@ -1,17 +1,60 @@
...
+    return max
+    } else {
+      throw new Error('EmptySortedList')
+    }
+  }
 
-  min() {}
+  min() {
+    if (this.length>0) {
+      var min = this.items.reduce(function(a, b) {
```
valmgisbert, line 34: Same thing as max! stick to one of these approaches!

index.js
```diff
@@ -1,17 +1,60 @@
...
+    const sum = this.items.reduce (function (accumulator, currentValue) {
+      return accumulator + currentValue
+    },0)
+    return sum
+  }
+  avg() {
+      if(this.length==0) {
+        throw new Error (`EmptySortedList`)
+      } else {
+        let totalAvg = (this.sum()/this.items.length).toFixed(2);
```
valmgisbert, line 53: nice in reusing your functions!

index.js
```diff
@@ -1,17 +1,60 @@
...
+  constructor() {
+    this.items = [],
+    this.length = 0
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort(function(a, b){return a-b});
+    this.length =+1;
```
valmgisbert, line 10: The proper syntax of the shorthand operators is the op symbol first, and then the equal (+=, -=...), otherwise it will throw an error.

index.js
```diff
@@ -1,17 +1,60 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [],
+    this.length = 0
```
valmgisbert, line 4: Okay, this is technically correct but ONLY for this case (an empty array); ideally, your length should adapt to whatever number it might have, with `this.items.length` here.

valmgisbert:
Nice work! Please check the comments and try to refactor a bit, you could make it work without certain steps.

PR #1319, opened 2020-07-28
Done by Marta and Bodei

MartaLourido:
done! very good lab, done in pairs :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,59 @@
 class SortedList {
-  constructor() {}
+  constructor(items, length) {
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Remember to use the code of your cohort in your pull requests: 
[WDFT-REM-2007] your name or names.
Regarding the lab, very good! all tests passing and looks clean, I would only recommend you to start using the short syntax for methods using arrow functions. Moving forward it will be better to use those.

PR #1356, opened 2020-09-01
Lab done with bonus Daniel P. and Sophia WebDev FT August 2020

SophiaFrederike:


SophiaFrederike:
avg Method wasn't committed:

 avg() {
    if (this.length === 0) {
      throw new Error('EmptySortedList')
    }else {
      return this.sum()/this.length;
    }

  }

PR #1380, opened 2020-09-15
[PAR-WDFT-248] LISA PONCET

lisapnct:


ta-web-paris:
Hey Lisa !
Good job on this lab !
Just one quick thing to make it perfect :
The test might have passed, but just logically since you didn't define methods like "throw new OutOfBounds( )" real code wouldn't work because the function would be undefined.

One good syntax for these would be :
throw new Error("OutOfBounds")

But if you had defined them it would work in real world !

Otherwise all good !

Nina

PR #1385, opened 2020-10-06
[ REM 2009 ] Lovro Hodak + Aleksandra Tanasijenko

alekstana:
Yahoo!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,65 @@
 class SortedList {
-  constructor() {}
+  constructor() {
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very good solutions, only one issue and it was that inside the avg method you have no access to the sumArr variable created in the sum method since those are different scopes. One way to run that part of the code (Lines 49 to 51) inside the avg method.
Still, overall well done. Keep it up!

PR #1386, opened 2020-10-06
[REM-2009] Sumiya Ushiro & Juan Jose

sumi0820:
We completed all iterations and bonus.


IH-WebDev-TA-Remote:
Very nice solutions, all iterations done and that use of Math.min and Math.max with the spread operator was perfect. Keep up the great work!

PR #1388, opened 2020-10-06
[REM-2009] Jordane Ruiz + Santiago Ardiles

SantAndresP:
;)


IH-WebDev-TA-Remote:
Very nice solutions and all iterations done! Good use of .reduce and reusing the same sum() method in the avg() method.
Keep up the great work!

PR #1447, opened 2020-11-05
Completed Mocha Test

MADFutcher:


index.js
```diff
@@ -1,17 +1,61 @@
...
+  }
 
-  max() {}
+  max() {
+    if(this.items.length === 0){
+      throw new Error('EmptySortedList');
+    }
 
-  min() {}
+    this.items.sort(function(a, b) {
```
In-S4n3, line 27: This is a very smart aproach! But you also can use the **Math.max()** inside a reduce method, take a look a it.

index.js
```diff
@@ -1,17 +1,61 @@
...
+      return b - a;
+    });
+    return this.items[0]
+  }
 
-  sum() {}
+  min() {
```
In-S4n3, line 33: Also, **Math.min()**

index.js
```diff
@@ -1,17 +1,61 @@
...
+      return 0
+    }
+    return this.items.reduce((a, b)=> a+b)
+  }
+
+  avg() {
+    if(this.items.length === 0){
+      throw new Error('EmptySortedList');
+    }
+    let total = this.items.reduce((a, b)=> a+b)
```
In-S4n3, line 55: Once you have the **sum()** already above, you could just divide it by the length 👍

PR #1448, opened 2020-11-05
[PTREMOTE092020]JoseAlbertoMontemayorGonzalez

Thrashskin:
Done

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,65 @@
 class SortedList {
-  constructor() {}
+   
+  constructor() {
+    this.items = [];
+    this.length = 0;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items = this.items.sort( (a,b) => {
+      return a-b;
+    });
+    this.length++;
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos < 0 || pos > this.items.length) {
+      throw new Error ('OutOfBounds');
+    } else {
+      return this.items[pos];
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.items.length === 0) {
+      throw new Error('EmtySortedList');
+    } else {
+      return Math.max(...this.items);
+    }
+  }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,17 +1,65 @@
...
+    this.items.push(item);
+    this.items = this.items.sort( (a,b) => {
+      return a-b;
+    });
+    this.length++;
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos < 0 || pos > this.items.length) {
```
In-S4n3, line 17: You could just aproach it this way = if(**this.items[pos]===undefined**)

index.js
```diff
@@ -1,17 +1,65 @@
...
+      return this.items[pos];
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.items.length === 0) {
+      throw new Error('EmtySortedList');
+    } else {
+      return Math.max(...this.items);
```
In-S4n3, line 28: here you can use two aproachs. A new one which is a Math.max() inside a reduce method (take a look at it) or simply sort the array and return the last or fist number depending of how you sort it.

Hope it made sense.

index.js
```diff
@@ -1,17 +1,65 @@
...
+    }
+  }
 
-  min() {}
+  min() {
+    if (this.items.length === 0) {
+      throw new Error('EmtySortedList');
+    } else {
+      return Math.min(...this.items);
```
In-S4n3, line 36: Here the same as above!

index.js
```diff
@@ -1,17 +1,65 @@
...
+    }
+  }
 
-  avg() {}
+  avg() {
+    let sum = 0;
+    if (this.items.length === 0) {
+      throw new Error('EmtySortedList');
+    } else {
+      sum = this.items.reduce((acc, elem) => {
```
In-S4n3, line 57: Once you already have the sum() function above you could just used it and divide it by the length.

PR #1449, opened 2020-11-06
[REMOTE-PT-092020] Dirk Kelderman 

dirkkelderman:


index.js
```diff
@@ -1,17 +1,56 @@
...
+    } else {
+      return this.items.indexOf(pos);
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.length === 0) {
+      throw TypeError("EmptySortedList");
+    } else {
```
In-S4n3, line 26: To return the maximum number inside an array, you can use this method **Math.max()**. Try it!

index.js
```diff
@@ -1,17 +1,56 @@
...
+      return this.items.slice(-1).pop()
+    }
+  }
 
-  min() {}
+  min() {
+    if (this.length === 0) {
+      throw TypeError("EmptySortedList");
+    } else {
+      return this.items[0];
```
In-S4n3, line 35: Well, the number in the index 0, can not be the smallest, has the example before, try use **Math.min()**

You're wellcome 👍

index.js
```diff
@@ -1,17 +1,56 @@
...
+      return this.items.reduce((a, b) => a + b, 0)
+    }
+  }
 
-  avg() {}
+  avg() {
+    if (this.length === 0) {
+      throw TypeError("EmptySortedList");
+    } else {
+      return (this.items.reduce((a, b) => a + b, 0) / this.length);
```
In-S4n3, line 51: You alredy did that above, you could just use the **sum()/length**.

PR #1450, opened 2020-11-07
[Remote-PT-092020]Zhenni QIAN

Jennie-QIAN:


index.js
```diff
@@ -1,17 +1,53 @@
...
+  }
 
-  max() {}
+  max() {
+    if (this.length === 0) {
+      throw new Error('EmptySortedList');
+    }
 
-  min() {}
+    return this.items[this.length - 1];
```
In-S4n3, line 28: Zhenni what if the highest number was in a random position? Take a look at the Math.max, or another solution could be by sorting the array and returning the first or the last depending of the sort! 👍

index.js
```diff
@@ -1,17 +1,53 @@
...
+  }
 
-  sum() {}
+  min() {
+    if (this.length === 0) {
```
In-S4n3, line 32: Same as above, take a look at Math.min()

PR #1461, opened 2020-11-08
[Remote-PTWD-092020] Daina Slezaite

daina-slezaite:


index.js
```diff
@@ -1,17 +1,57 @@
...
+  }
+  get(pos) {
+    if(pos > this.items.length) {
+      throw new Error('OutOfBounds');
+    } else{
+      return this.items[pos];
+    }
+  }
+  max() {
+    if(this.items.length === 0) {
```
In-S4n3, line 19: Daina what if the highest number was in a random position? Take a look at the Math.max, or another solution could be by sorting the array and returning the first or the last depending of the sort! 👍

index.js
```diff
@@ -1,17 +1,57 @@
...
+      throw new Error('EmptySortedList');
+    } else{
+      let i = this.items.length - 1;
+      return this.items[i];
+    }
+  }
+  min() {
```
In-S4n3, line 26: Same as above, take a look at the Math.min()

index.js
```diff
@@ -1,17 +1,57 @@
...
+      return acc + current;
+      })
+      return sum;
+    }
+  }
+  avg() {
+    if(this.items.length === 0) {
+      throw new Error('EmptySortedList');
+    } else {
+      let total = 0;
```
In-S4n3, line 48: Once you already calculated the sum() above, you could just use it here and divide it by the length.

PR #1464, opened 2020-11-09
all iterations done

ZuzanaAK:


index.js
```diff
@@ -1,17 +1,55 @@
...
-  avg() {}
+  sum() {
+      if (this.length != 0) {
+          return this.items.reduce((accu, curr) => accu + curr);
+      } else {
+          return 0;
+      }
+  }
+
+  avg() {
```
In-S4n3, line 46: Once you calculated above the sum() you could just divide it from the length :)

index.js
```diff
@@ -1,17 +1,55 @@
...
+  get(pos) {
+      if (this.items[pos] != undefined) {
+          return this.items[pos];
+      } else {
+          throw new Error('OutOfBounds');
+      }
+  }
 
-  max() {}
+  max() {
```
In-S4n3, line 21: Very nice!

index.js
```diff
@@ -1,17 +1,55 @@
...
+      if (this.length != 0) {
+          return Math.max(...this.items);
+      } else {
+          throw new Error('EmptySortedList');
+      }
+  }
 
-  min() {}
+  min() {
+      if (this.length != 0) {
```
In-S4n3, line 30: Very nice!

PR #1483, opened 2020-11-10
[WDFTRM-20201110] - Hugo Embleton-Black & Pauline Massé

paulinemss:


filipe-freire (commented):
Very nicely done! Getting used to ES6 syntax I see 😎 Nothing more to add here! 🥇

PR #1486, opened 2020-11-10
[FTRMTWD2020] - Marcel & Alex

mnblski:
all done

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,62 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort((a, b) => a - b);
+    this.length = this.items.length;
+  }
 
-  get(pos) {}
+  get(pos) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,17 +1,62 @@
...
+  }
 
-  sum() {}
+  sum() {
+    let sum = 0;
 
-  avg() {}
+    for (let i = 0; i < this.length; i++) {
+      sum += this.items[i];
+    }
```
filipe-freire, line 42: You could also have used an array method here! Try using `.reduce` to achieve the same goal. 💪

index.js
```diff
@@ -1,17 +1,62 @@
...
+
+    return sum;
+  }
+
+  avg() {
+    let sum = 0;
+
+    for (let i = 0; i < this.length; i++) {
+      sum += this.items[i];
+    }
```
filipe-freire, line 52: Since we already declare a sum method above, we could call it instead of rewriting it here! Remember the DRY principle! 😉

filipe-freire (commented):
Good work overall! Left a few comments that improve on the quality of your code, but very nice job nonetheless. 👍

PR #1489, opened 2020-11-10
[WDFTRM - 2020-11-10] - Dimitri and Gosia

GosiaMas:


filipe-freire (commented):
Awesome work here! Bonus points for using ES6 syntax! 🎖

PR #1490, opened 2020-11-10
[WDFTRM-20201011] Querien Ornstein & Marc Gotzens

Mgot12602:


index.js
```diff
@@ -1,17 +1,56 @@
...
+  }
 
-  min() {}
+  sum() {
+    if (this.items.length === 0) {
+      return 0;
+    }
+    let sum = this.items.reduce(function (acc, el) {
+      return acc + el;
+    }, 0);
```
filipe-freire, line 41: Since we'll be using ES6 syntax a lot from now on, the reduce method could also be written like this: `.reduce((acc, el) => acc + el, 0);`

index.js
```diff
@@ -1,17 +1,56 @@
...
-  sum() {}
-
-  avg() {}
+  avg() {
+    if (this.items.length === 0) {
+      throw new Error("EmptySortedList");
+    }
+    let sum = this.items.reduce(function (acc, el) {
+      return acc + el;
+    }, 0);
```
filipe-freire, line 51: Since we already defined a sum method above, we could also call it here by writing `this.sum()`. Remember the DRY principle! ✔

filipe-freire (commented):
Lab nicely done! Good job 😁

PR #1497, opened 2020-11-13
[REMOTE-PT-092020] Ashley Vermaas

ashleyvermaas:
- Still need to refactor
- Couldn't find out why it wasn't passing this test: "should add a single value to SortedList"
- Need to figure out how to make the 'sum' variable accessible in other methods such as avg()
- Need to find out why tests wouldn't pass when 'this.length' was used instead of 'this.items.length'

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,61 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+    this.items.sort(function (a, b) {
+      return a - b;
+    });
+  }
 
-  get(pos) {}
+  get(pos) {
+    if (pos <= this.items.length && pos >= 0) {
+      var value = this.items[pos];
+      return value;
+    } else {
+      throw new Error('OutofBounds');
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.items.length > 0) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,17 +1,61 @@
...
+      var sum = this.items.reduce((a, b) => a + b);
+      return sum;
+    } else {
+      return 0;
+    }
+  }
 
-  avg() {}
+  avg() {
+    if (this.items.length > 0){
```
In-S4n3, line 51: You just calculated the sum() above, you could just divite it by the length :)

index.js
```diff
@@ -1,17 +1,61 @@
...
+      var value = this.items[pos];
+      return value;
+    } else {
+      throw new Error('OutofBounds');
+    }
+  }
 
-  max() {}
+  max() {
+    if (this.items.length > 0) {
```
In-S4n3, line 24: You could have two other aproaches here:
1) Sort everthing and return the first or last value depending if you sorte ascending or descending.
2) Math.max(), take a look a this method.

index.js
```diff
@@ -1,17 +1,61 @@
...
+      var highestValue = this.items[this.items.length - 1];
+      return highestValue;
+    } else {
+      throw new Error('EmptySortedList');
+    }
+  }
 
-  min() {}
+  min() {
+    if (this.items.length > 0) {
```
In-S4n3, line 33: Same as above, take a look at the Math.min()

PR #1611, opened 2021-04-12
[FTWDAMS032021] Igor Godinho Figueiredo

Igor-GF:


index.js
```diff
@@ -1,17 +1,59 @@
...
+    }
+  }
 
-  avg() {}
+  sum() {
+    if(this.items) {
+      return this.items.reduce((el1, el2) => el1 + el2);
+    } else {
+      return 0;
+    }
```
reMRKableDev, line 51: Consider using the length of the array in your condition here instead.

```suggestion
    if(this.items.length) {
      return this.items.reduce((el1, el2) => el1 + el2);
    } else {
      return 0;
    }
```

reMRKableDev (commented):
Nice attempt @Igor-GF 🚀 

There are some conceptual things you can revisit here. I've left some comments and suggestions for you to check out

PR #1850, opened 2022-01-22
[MDPT0122] MIGUEL BAÑOS

miguelbanos95:
lab done + bonus

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,57 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = []
+    this.length = this.items.length
+  }
 
-  add(item) {}
+  add(item) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,17 +1,57 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = []
+    this.length = this.items.length
+  }
 
-  add(item) {}
+  add(item) {
+    if (this.length = 1) {
```
numbing, line 8: hello from berlin :) i was just checking the PR's and i see here you write only one "=" that and assignment and not a check and for checking if this.length is = to 1, you have to write "=="
have a nice time in sunny SP :)

PR #1933, opened 2022-05-27
[RMTEU-WDPT-APR2022] Tanvi Gurjar

Tanz1712:


index.js
```diff
@@ -1,17 +1,42 @@
...
+    const sum = this.items.reduce((a, b) => a + b, 0);
+    return sum;
+  }
 
-  avg() {}
+  avg() {
+    if(!this.length) {
+      throw new Error('EmptySortedList');
+    }
+    const avg = this.items.reduce((a, b) => a + b, 0) / this.length;
```
Matteo-Ironhack, line 37: if an average is "sum over length", you could use the sum method. This also work, but you have code repetition

Tanz1712:
Yes i had tried but got some error, so had to repeat the code

PR #1935, opened 2022-05-27
lab-intro-node

zarahsara:


index.js
```diff
@@ -1,13 +1,39 @@
...
+  }
 
-  add(item) {}
+  add(item) {
+    this.items.push(item);
+     this.length++;
+   return this.items.sort(function(a, b){
+   return a - b;
+  })
+   return this.items;
```
Matteo-Ironhack, line 14: this `return` statement will never be called because you have one before this line in the same function

index.js
```diff
@@ -1,13 +1,39 @@
 class SortedList {
```
Matteo-Ironhack, line 1: remember to check for syntax errors. Since one parenthesis went missing, nothing works (even if the logic is good)

index.js
```diff
@@ -1,13 +1,39 @@
...
+    this.items.push(item);
+     this.length++;
+   return this.items.sort(function(a, b){
+   return a - b;
+  })
+   return this.items;
+}
+
+  get(pos) {
+    if (pos.this.length && !this.length){
```
Matteo-Ironhack, line 18: pos has no attribute `this`. You should check if `pos` (a number) is lesser than the items array's length

PR #1936, opened 2022-05-28
 lab-intro-node

ihab5520:


index.js
```diff
@@ -1,17 +1,90 @@
...
+  get(pos) {
+    if (!this.items[pos]) {
+      throw new Error('OutOfBounds');
+    }
+    return this.items[pos];
+  }
 
-  max() {}
+  max() {
+    if (this.items.length === 0) {
```
Matteo-Ironhack, line 26: this works, but remember that they also made you create a `length `property in the constructor

index.js
```diff
@@ -1,17 +1,90 @@
...
+        number = item;
+      }
+    });
+    return number;
+  }
 
-  avg() {}
-}
 
+  sum ( numbers )
```
Matteo-Ironhack, line 56: `sum` takes no arguments

index.js
```diff
@@ -1,17 +1,90 @@
...
+  {
+	  let sum = 0; 
+	  for (let i = 0; i < numbers.length; i++) {
+	    if (typeof numbers[i] === "number") {
```
Matteo-Ironhack, line 60: if you take a look at the test spec, you'll see you only need to handle numbers

index.js
```diff
@@ -1,17 +1,90 @@
...
+	      sum += numbers[i];
+	    } else {
+	      throw new Error("err");
+	    }
+	  }
+	  return sum; 
+	}
+
+
+  avg(numbersAvg) {
```
Matteo-Ironhack, line 74: `avg` takes no arguments

PR #1937, opened 2022-05-28
solved with problems

EvgeniiaRiaguzova:


index.js
```diff
@@ -1,16 +1,55 @@
...
+    this.items=[];
+    this.length= this.items.length;
+  }
 
-  add(item) {}
+  add(item) {
+    this.item=item;
+    this.items.push(item);
+    this.length +=1;
+    this.items.sort((a,b)=>{a-b});
```
Matteo-Ironhack, line 11: good logic, but the function called by `sort` should **return** the value (in this case `a - b`) for the function to work properly. Otherwise the array will not be sorted

index.js
```diff
@@ -1,16 +1,55 @@
...
-  get(pos) {}
+  get(pos) {
+    if (pos>=this.items.length) 
+    {throw new Error('OutOfBounds')}
+    else {return this.items[pos]}
+  }
 
-  max() {}
+  max() {
+    if (this.items=[])
```
Matteo-Ironhack, line 21: with just one `=` sign you are assigning a new value to the variable (in this case, you are setting `this.items` to an empty array every time I call the `max` function

Matteo-Ironhack: if you already have a sorted array (because of the add method), you can just return the last element

PR #1938, opened 2022-05-28
solved lab

nataliabuzila:


index.js
```diff
@@ -1,17 +1,51 @@
 class SortedList {
-  constructor() {}
+  constructor() {
+    this.items = [];
+    this.length = this.items.length;
+  }
 
-  add(item) {}
+  add(item) {4
```
Matteo-Ironhack, line 7: careful, the `4` you left there could make all the tests fail. In this case it does not, but remeber to check for typos

index.js
```diff
@@ -1,17 +1,51 @@
...
-  get(pos) {}
+  get(pos) {
+    if(pos>this.items.length) throw new Error('OutOfBounds');
+    return this.items[pos];
+  }
 
-  max() {}
+  max() {
+    if(!this.items.length) throw new Error('EmptySortedList');
+    let maxValue = 0;
```
Matteo-Ironhack, line 20: if the array is sorted, then you don't need to check every item for the biggest number; you know where it is

